Issue #17, opened 2016-03-02
Build using latest gherkin-go is broken


odacremolbap:
2 days ago `gherkin-go` updated some types declared at `ast.go`

ScenarioOutline.Examples is now an `[]interface{}`
https://github.com/cucumber/gherkin-go/blob/master/ast.go#L46-L50
```
type ScenarioOutline struct {
	ScenarioDefinition
	Tags     []*Tag        `json:"tags"`
	Examples []interface{} `json:"examples"`
}
```
Code that uses Examples is now failing due to not recognised methods.

**Scenario:**

```
godog$ git rev-parse HEAD
38c79db503ff4ad0cbb7c1fd04531f243d992c7d
```

```
gherkin-go$  git rev-parse HEAD
64f5ebccd5511abba41bf4fa5664faf289c9b102
```

go versions tested
osx (1.4.3, 1.5.0, 1.6.0)
alpine docker container (1.4.3)

**Build output:**
```
/go/src/github.com/DATA-DOG/godog # go build .
# github.com/DATA-DOG/godog
./fmt.go:224: ex.TableBody undefined (type interface {} has no field or method TableBody)
./fmt_pretty.go:136: example.TableBody undefined (type interface {} has no field or method TableBody)
./fmt_pretty.go:176: example.TableHeader undefined (type interface {} has no field or method TableHeader)
./fmt_pretty.go:181: example.Keyword undefined (type interface {} has no field or method Keyword)
./fmt_pretty.go:181: example.Name undefined (type interface {} has no field or method Name)
./fmt_pretty.go:183: example.TableHeader undefined (type interface {} has no field or method TableHeader)
./fmt_pretty.go:190: example.TableBody undefined (type interface {} has no field or method TableBody)
./suite.go:271: example.TableHeader undefined (type interface {} has no field or method TableHeader)
./suite.go:272: example.TableBody undefined (type interface {} has no field or method TableBody)
./suite.go:272: too many errors
```

l3pp4rd:
indeed, will need to adapt it, thanks for the update

odacremolbap:
I'm sending a PR in a few seconds to add  ```godep``` linked to the previous gherkin-go commit.
Testing right now.

l3pp4rd:
the master was updated with gopkg.in release tag based package dependency. will need to follow it up with gherkin-go to next releases whether it will follow semver or not

PR #18, opened 2016-03-02
add godep to use dependencies/revision

odacremolbap:
Add godep, linking gherkin-go on 0c0f2d933220a0ebc0dd5e35d7c73dd5b4968d85 commit to workaround #17 

Test dependencies not included automatically in godep.
```/github.com/go-sql-driver/mysql```should be added manually


odacremolbap:
Travis is failing because it's executing `go get github.com/cucumber/gherkin-go`
I'm fixing that one bringing godep to `.travis.yml`

l3pp4rd:
as a temporary solution I'd rather change it to gopkg.in/cucumber/gherkin-go.v3 which would explicitly require a release version, instead of introducing godeps. gherkin-go may make a release under v3 with backward incompatible changes, in that case adapting source code would be necessary

odacremolbap:
Great. I'll take a look into it, but I don't think I have time to change that at this moment.

odacremolbap:
OK, it's working now.
Thanks!

PR #36, opened 2016-06-08
Adding -f=text option to provide colorless output

lexandro:
I'd like to add a pretty, but uncolorized output to the godog tool, because not all CI tool can handle ANSI colors by default.

### DIFF
--- a/fmt_text.go
+++ b/fmt_text.go
@@ -1,0 +1,362 @@
+package godog
+
+import (
+	"fmt"
+	"math"
+	"strings"
+	"time"
+	"unicode/utf8"
+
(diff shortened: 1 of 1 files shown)


l3pp4rd:
this needed to be done in more practical way, but thanks. you could have opened an issue before to discuss the way you are going to implement it. Regarding CI server, I suggest you run with progress format: `godog -f progress --no-colors` less clumsy output, plus it prints error messages for failed scenarios, basically only useful information in ten lines.

lexandro:
Thx for the advice, but the progress format is not enough. There's a reason why I demanded a colorless output.

lexandro:
btw thx. :)

PR #43, opened 2016-06-22
use maximum step length of background and scenario

nightlyone:
in order to avoid panic in printStep (line 217).

Fixes #42

PR #45, opened 2016-06-27
Adding nil ref check

### LINKED ISSUE #44: Panic in fmt_pretty
Probably related to #42

Attempting to run godog in the godog code folder results in the following panic:

image

OwenGr:
This is only my third day on Go, so this may not be the best way to fix this. Basically, #43 doesn't do a nil reference check before calling `f.length(f.feature.Background)`. I don't know if this should have been resolved further up the chain. This resolves #44.


l3pp4rd:
ah, it was introduced recently. Well thanks a lot, was on vacation, thought to write some tests. anyway, both should be solved now, thanks again for your contributions.

PR #48, opened 2016-07-14
Updated API example to now use FeatureContext func name rather than f…

pmcgrath:
Updated API example to now use FeatureContext func name rather than featureContext, also updated the screenshots

If someone runs 
```
godog .
```
It will indicate they need to rename the function, but its always good when an example just works


l3pp4rd:
thanks ;)

PR #55, opened 2016-10-27
Build executable - new command line switch -o

edharper01:
This change adds support for a new switch (-o) which is used to generate a copy of the test runner executable instead of running any tests.
This allows features to be run in environments where a full Go installation is not available.

Having struggled to find a suitable way to provide a unit test for this feature I'm raising the pull request in the hope that smarter minds than mine might have a suggestion.


l3pp4rd:
thanks, I will review that later ;)

l3pp4rd:
thanks, I've pulled it and merged, just had to run `go fmt` and rebase because you had this pull from master merge commit. everything is now in master, have fun ;)

PR #57, opened 2016-12-20
New: possibility to print delayed messages in an external lib or in your tests

Dogild:
Previously, it was not possible to print messages from an external library in godog.
Actually, it was possible, but the logs would be printed before the steps. 
- This PR adds the possibility to print these messages after the steps, this helps a lot when debugging a tests or you code.
- This PR adds the possibility to create easily external formatter as well, an example has been added in the README.
- The interface Formatter now needs to implement the method `Output()`, this method needs to return the io.writer used by the current formatter. The basent struct returns now a new struct `godog.NewDelayedMessagesWriter`, this allows the formatter to print the messages in a good order. - This is kind of similar to how the cucumber ruby version works with delayed messages.
- When running in concurrency, each suite has now its own formatter, this is way simpler to handle with concurrency, however global summary has been broken...it works as in parallel_test (ruby lib)
- DelayedMessage has only activate in the pretty format though!

### DIFF
--- a/README.md
+++ b/README.md
@@ -246,13 +246,19 @@ func minimaleFunc(suite string, out io.Writer) godog.Formatter {
 		out:     out,
 		indent:  2,
 		started: time.Now(),
+    delayedMessageWriter: godog.NewDelayedMessagesWriter(),
 	}
 }
 
 type minimale struct {
+  delayedMessageWriter *godog.DelayedMessagesWriter
 	out     io.Writer
 	indent  int
 	started time.Time
+}
+
+func (m *minimale)Output() io.writer {
+  return m.delayedMessageWriter
 }
 
 func (m *minimale) Feature(f *gherkin.Feature, s, string, b, []byte) {
@@ -295,10 +301,16 @@ func (m *minimale) Summary() {
 }
 ```
 
-Then make sure to import this package in your mail file :
+Then make sure to import this package in your main file:
 
 ```go
 import _ "path"
+```
+
+Now to print any messages in your tests or external library:
+
+```go
+fmt.Fprintln(c.suite.Output(), strings.Repeat(" ", 10)+colors.Cyan("my log"))
 ```
 
 ### References and Tutorials
--- a/README.md
+++ b/README.md
@@ -228,6 +228,79 @@ func TestMain(m *testing.M) {
 }
 ```
 
+### Usage of a custom formatter
+
+You can implement a custom format when running godog. To achieve that, you need to implement a struct which implement the interface Formatter.
+
+See the following example :
+
+```go
+package fmtminimale
+
+func init() {
+	godog.Format("minimale", "Prints minimal status", minimaleFunc)
+}
+
+func minimaleFunc(suite string, out io.Writer) godog.Formatter {
+	return &minimale{
+		out:     out,
+		indent:  2,
+		started: time.Now(),
+	}
+}
+
+type minimale struct {
+	out     io.Writer
+	indent  int
+	started time.Time
+}
+
+func (m *minimale) Feature(f *gherkin.Feature, s, string, b, []byte) {
+  fmt.Println("[FEATURE]" + f.Name)
+}
+
+func (m *minimale) Node(node interface{}) {
+  switch t := node.(type) {
+  case *gherkin.Scenario:
+    fmt.Println("[SCENARIO]" + t.Name)
+  }
+}
+
+func (m *minimale) Defined(s *gherkin.Step, sd *godog.StepDef) {
+  fmt.Println("[DEFINED]" + s.Name)
+}
+
+func (m *minimale) Failed(s *gherkin.Step, sd *godog.StepDef, e error) {
+  fmt.Println("[FAILED]" + s.Name)
+}
+
+func (m *minimale) Passed(s *gherkin.Step, sd *godog.StepDef) {
+  fmt.Println("[PASSED]" + s.Name)
+}
+
+func (m *minimale) Skipped(s *gherkin.Step) {
+  fmt.Println("[SKIPPED]" + s.Name)
+}
+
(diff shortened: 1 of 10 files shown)


l3pp4rd:
thanks, I'll review it in few days

cmd/godog/main.go
```diff
@@ -36,6 +36,7 @@ func buildAndRun() (int, error) {
 	cmd := exec.Command(bin, os.Args[1:]...)
 	cmd.Stdout = os.Stdout
 	cmd.Stderr = os.Stderr
+	cmd.Stdin = os.Stdin
```
l3pp4rd: godog does not accept stdin, there is no point to have it in tests

README.md
```diff
@@ -228,6 +228,91 @@ func TestMain(m *testing.M) {
...
+	out     io.Writer
+	indent  int
+	started time.Time
+}
+
+func (m *minimale)Output() io.writer {
+  return m.delayedMessageWriter
+}
+
+func (m *minimale) Feature(f *gherkin.Feature, s, string, b, []byte) {
```
l3pp4rd: there are many mistakes in this code sample. It is too simple to read the godoc and implement the formatter if they need it. so far everyone was able to manage it without questions.

README.md
```diff
@@ -228,6 +228,91 @@ func TestMain(m *testing.M) {
...
+func init() {
+	godog.Format("minimale", "Prints minimal status", minimaleFunc)
+}
+
+func minimaleFunc(suite string, out io.Writer) godog.Formatter {
+	return &minimale{
+		out:     out,
+		indent:  2,
+		started: time.Now(),
+    delayedMessageWriter: godog.NewDelayedMessagesWriter(),
```
l3pp4rd: you do not follow formatting rules, please use `go fmt`

README.md
```diff
@@ -228,6 +228,91 @@ func TestMain(m *testing.M) {
...
+}
+
+type minimale struct {
+  delayedMessageWriter *godog.DelayedMessagesWriter
+	out     io.Writer
+	indent  int
+	started time.Time
+}
+
+func (m *minimale)Output() io.writer {
```
l3pp4rd: again code style formatting

flags.go
```diff
@@ -33,6 +33,7 @@ func FlagSet(opt *Options) *flag.FlagSet {
 	for name, desc := range AvailableFormatters() {
 		descFormatOption += s(4) + "- " + colors.Yellow(name) + ": " + desc + "\n"
 	}
+	descFormatOption += s(4) + "- " + colors.Yellow("<format>") + ": name of your custom format" + "\n"
```
l3pp4rd: godog does not compile package on `help` otherwise it would show the custom format in the list. probably later we will make it to compile the package, since after 1.5 release of go compilation was slower, now it is getting faster again, probably after go 1.8 this will show the custom formats in `godog -help`

fmt.go
```diff
@@ -90,6 +93,46 @@ func AvailableFormatters() map[string]string {
 	return fmts
 }
 
+// DelayedMessagesWriter is a structure used to print messages from a external library
+// This struct is useful when printing logs from the test themself
+// It will guarantee the orders of the log
+// For instance, when using this struct, the format will first print the step name and then all of the logs printed in this step
+type DelayedMessagesWriter struct {
```
l3pp4rd: I do not see agree with this one. If you have read the [contribution guide](https://github.com/DATA-DOG/godog#contributions) in the end of the readme. you should have noticed that changes and additions to public interface are and will be treated carefully. I will not accept these unless they are really useful changes.

In this case, you are contributing a change which is not necessary. maybe it makes it easier for user to do it, but it adds redundancy to public godog interface and in time such useless features will only confuse people and inspire bad practices.

See the comment on main pull request, explaining how you should implement such behavior, because godog already gives you these options.

l3pp4rd:
Here is the [commit](https://github.com/DATA-DOG/godog/commit/219f15bd92f513e9bbbfc8af08232caffdcbd539) I made to show you how you should approach this. there is an [after step](https://godoc.org/github.com/DATA-DOG/godog#Suite.AfterStep) hook, which allows you to print such delayed messages, it gives more, the ability to print different messages or test suite context in case of an error after step was executed.

Dogild:
No problem, you solution will work as well, I have proposed this PR to make godog reacts like cucumber ruby for the messages, but it does not matter if you think that the developer should handle that! I'll clean the PR or close it actually.

There is one other feature which was added to the PR, the `cmd.Stdin = os.Stdin`. I have proposed that because in one of mine personal formatter, I need stdin ; I have a formatter where each steps are paused till the user press the key enter. Since the cmd does not have the stdin associated to godog, it won't work without this path. There is no problem if you don't accept it in godog, I can use my fork and sync with the master sometimes, I just let you know about this use case. Let me know if I need to update this PR, or if you believe it's not mandatory to godog, just close the PR ;)

l3pp4rd:
Concerning Stdin, I do not think this is a good idea at all. If you have your godog features depending on stdin input at some point, then you cannot run it automatically in any continues integration server. export these to environment variables or compile based tags if you have security related stuff. Also I do not think that any configuration could be accepted from stdin for tests.

But in general probably we can pass in this stdin input, strip it down to only os.Stdin assignment to command and that is it, though I do not think of right reasons to use it, but it makes sense to pass it on anyway.

Dogild:
Yes I know for the stdin and the integration system. We run our features with a custom formatter on concourse which does not need stdin. But once a test fails, we give the options to the developer to run the test locally on his system, with an option he can then put a pause between each tests, way easier to debug when having long scenarios

Dogild:
I closed this PR, the PR #59 and #60 implements some features created in this PR

l3pp4rd:
well, again, use the mutex. I will not accept any extra features which can be easily implemented in your feature suite.

PR #60, opened 2017-01-05
Fixed: do not exit godog if the given format is not a godog format and create a different formatter for each suite for concurrency

Dogild:
Fixed: do not exit godog if the given format is not a godog format and if the concurrency is greater than 1. 

Fixed: We know create a formatter for each suite when launching godog in concurrency

### DIFF
--- a/fmt.go
+++ b/fmt.go
@@ -37,6 +37,8 @@ var undefinedSnippetsTpl = template.Must(template.New("snippets").Funcs(snippetH
 }
 `))
 
+var godogFormats = []string{"events", "junit", "pretty", "progress"}
+
 type undefinedSnippet struct {
 	Method   string
 	Expr     string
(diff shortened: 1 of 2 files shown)


l3pp4rd:
sorry about the delay, very busy these days. I'll come back for the review then I have time for it

l3pp4rd:
it now allows custom formatters to run concurrently. But I cannot accept different formatter instances in concurrent mode, because each would print summary after each feature, this is not the intention of concurrency support. (single formatter should collect all summary)

If you want to run it the way you proposed (summary for each feature file) make a shell script or a golang program which finds all feature files and runs them each separate with `godog -f anyformat my.feature` and collect the status results exit if one of the runs failed with an error.

I do not intend to wrap all possible use cases in a godog command, you have all available tooling to run it anyway you like. follow the unix philosophy and compose

PR #62, opened 2017-01-16
New: added method FailWithError.

Dogild:
This method allows you to fail a suite whenever you want, for instance you could fail a suite of tests in a hook!

Not sure if it's the best way to do that, I'm listening if you have a better approach!

### DIFF
--- a/suite.go
+++ b/suite.go
@@ -48,8 +48,9 @@ type Suite struct {
 	features []*feature
 	fmt      Formatter
 
-	failed        bool
-	stopOnFailure bool
+	arbitraryFailureError error
+	failed                bool
+	stopOnFailure         bool
 
 	// suite event handlers
 	beforeSuiteHandlers    []func()
@@ -163,11 +164,22 @@ func (s *Suite) AfterSuite(f func()) {
 	s.afterSuiteHandlers = append(s.afterSuiteHandlers, f)
 }
 
+// FailWithError makes fail the suite with the given error
+func (s *Suite) FailWithError(err error) {
+	s.arbitraryFailureError = err
+	s.failed = true
+}
+
 func (s *Suite) run() {
 	// run before suite handlers
 	for _, f := range s.beforeSuiteHandlers {
 		f()
 	}
+
+	if s.failed {
+		return
+	}
+
 	// run features
 	for _, f := range s.features {
 		s.runFeature(f)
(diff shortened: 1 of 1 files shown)


l3pp4rd:
I will leave it hanging, if more than a few users will express a need for such a feature, then it could be implemented, but differently - by returning an `ErrFatal` in a step, instead of having a method in Suite. 

the suite provides you with an option `--stop-on-failure` that should be enough, it also allows you to run a single scenario in case if you need to rerun it. You do not have to run the whole suite. this method confuses, user might think that is a preferred way. Please do not, make pull requests just for the sake of it. Any additional public method just adds garbage and complexity to understand the overall interface available to an user. I would rather look how to throw something away instead of adding.

Dogild:
By returning `ErrFatal` or using the option `--stop-on-failure` in one of the steps will not fixt the purpose of this PR.

I have proposed this PR to have the possibility to make fail a test from a hook (use case, you check if the env is clean at the end of the test, if not the test is a failure, you could panic but you don't want that). It's not something new for the user of cucumber as it exists exactly the same one in ruby-cucumber.

No comment about making a PR or not, be glad that some people wants to contribute, and maybe if someone needs something, someone else will need it as well. Purpose of the PR

l3pp4rd:
I do not agree with such feature in general, because this is a test suite, not a production server, if something is terribly wrong with a test context, then you should fail fast and dump this context in order to fix your tests for good, if you need such a feature - most probably that means you have an already too complex suite or maybe you have false positive tests (which is wrong in the first place). The point is, that your test suite is not an application, it is meant to show you whether your feature scenarios pass or fail. if they are wrongly implemented, then implement them well, that is the end of the story. Cucumber ruby, has many things, which I would not like to have in godog. Adding a feature is easy - removing it is hard, I want to be certain that this would pay off. How long do your tests run? If you use database, use transactions and roll them back on each scenario. If your suite is running longer than a minute, refactor.

Dogild:
We use godog for functional end-to-end testing, where our QA guys don't know how the system is working internally and write story for the product.

For the database story, they don't have access to it, as they have the product as a black box, only the public API are exposed. 

We would like to add this kind of test in the hook to make sure that the tester does not forget to clean everything (because it often happens unfortunately...). We know that it's not a good practice in terms of pure testing, but it helps a lot to control what is going on in the test. And we make sure to have standalone tests at the end.

l3pp4rd:
I can't add this in general because there is a concurrency option, which runs not only one suite in parallel. If I add a similar feature, I have to remove other, which I think is better.

I do not think it is wise to make end to end tests this way, in general you are making your application to support black box tests, even though, in most cases data should not be removed through public api endpoints.

I would recommend to use a fork or a before scenario hook to always clean the state before any scenario, without even trying to assert.

PR #67, opened 2017-03-01
Added support for the cucumber json format.

ckstettler:
I have added a format called "cucumber" which represents the cucumber json format.
I included cucumber.feature file to test and updated existing tests to fix issue associated with introducing the new feature file.

My least favorite part of my code is in suite_test.go where the matching logic used for the cucumber testing.  Certain fields like duration and match.location were problematic and you couldnt do a straight compare.  I had to create functions to compare the structure to make sure mutual (expected vs actual) field values matched plus no structure contained more fields than the other.

### DIFF
--- a/features/formatter/cucumber.feature
+++ b/features/formatter/cucumber.feature
@@ -1,0 +1,551 @@
+Feature: cucumber json formatter
+  In order to support tools that import cucumber json output
+  I need to be able to support cucumber json formatted output
+
+  Scenario: Support of Feature Plus Scenario Node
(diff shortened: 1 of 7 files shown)


l3pp4rd:
Hi, I think you forgot to commit `features/formatter/cucumber.feature`

ckstettler:
Yes I did. It is all there now.  My apologies.

l3pp4rd:
I'll review fully when I can, currently a busy week, thanks

PR #71, opened 2017-04-12
Create test01.feature

timdong:
test


l3pp4rd:
Hi, is this something I should care about? the feature file needs at least:

``` gherkin
Feature: describes something
```

PR #76, opened 2017-04-25
allow randomizing scenario order

mroth:
*WORK IN PROGRESS FOR DISCUSSION, NOT READY TO MERGE*

## Summary

Allow running scenarios in random order.

> This is especially helpful for detecting situations where you have state leaking between scenarios, which can cause flickering or fragile tests.

Similar to: https://www.relishapp.com/cucumber/cucumber/docs/cli/randomize

## Discussion
This isn't quite ready, but I wanted to open it for discussion with the maintainers prior to doing too much work on it.

## TODO
If this was to be adopted, the following would likely need to be done at minimum before merging:

1. Add some sort of `--random` flag to CLI app (will leave it up to maintainers to decide what the ideal name would be?)
2. update `--help` documentation to reflect/explain this flag.

Additionally, the following might be desirable:

1. Print the random seed at the end of a test run
2. Allow random seed to be specified via a command line `--flag` of some sort
3. Integration tests? (not quite sure how that would be implemented)

### DIFF
--- a/options.go
+++ b/options.go
@@ -13,6 +13,13 @@ type Options struct {
 	// Print step definitions found and exit
 	ShowStepDefinitions bool
 
+	// Run scenarios in random order.
+	//
+	// This is especially helpful for detecting situations
+	// where you have state leaking between scenarios, which
+	// can cause flickering or fragile tests.
+	RandomOrder bool
+
 	// Stops on the first failure
 	StopOnFailure bool
 
--- a/options.go
+++ b/options.go
@@ -13,16 +13,22 @@ type Options struct {
 	// Print step definitions found and exit
 	ShowStepDefinitions bool
 
-	// RandomSeed, if not `0`, will be used to run scenarios in a random order.
+	// Randomize causes scenarios to be run in random order.
 	//
 	// Randomizing scenario order is especially helpful for detecting
 	// situations where you have state leaking between scenarios, which can
 	// cause flickering or fragile tests.
+	Randomize bool
+
+	// RandomSeed allows specifying the seed to reproduce the random scenario
+	// shuffling from a previous run.
 	//
-	// The default value of `0` means "do not randomize".
+	// When `RandomSeed` is left at the nil value (`0`), but `Randomize`
+	// has been set to `true`, then godog will automatically pick a random
+	// seed between `1-99999` for ease of specification.
 	//
-	// The magic value of `-1` means "pick a random seed for me", the resulting
-	// seed will only be between `1-99999` for ease of specification.
+	// If RandomSeed is set to anything other than the default nil value (`0`),
+	// then `Randomize = true` will be implied.
 	RandomSeed int64
 
 	// Stops on the first failure
--- a/options.go
+++ b/options.go
@@ -13,12 +13,17 @@ type Options struct {
 	// Print step definitions found and exit
 	ShowStepDefinitions bool
 
-	// Run scenarios in random order.
+	// RandomSeed, if not `0`, will be used to run scenarios in a random order.
 	//
-	// This is especially helpful for detecting situations
-	// where you have state leaking between scenarios, which
-	// can cause flickering or fragile tests.
-	RandomOrder bool
+	// Randomizing scenario order is especially helpful for detecting
+	// situations where you have state leaking between scenarios, which can
+	// cause flickering or fragile tests.
+	//
+	// The default value of `0` means "do not randomize".
+	//
+	// The magic value of `-1` means "pick a random seed for me", the resulting
+	// seed will only be between `1-99999` for ease of specification.
+	RandomSeed int64
 
 	// Stops on the first failure
 	StopOnFailure bool
--- a/options.go
+++ b/options.go
@@ -2,9 +2,6 @@ package godog
 
 import (
 	"io"
-	"math/rand"
-	"strconv"
-	"time"
 )
 
 // Options are suite run options
@@ -33,7 +30,7 @@ type Options struct {
 	// Any other value will be used as the random seed for shuffling. Re-using the
 	// same seed will allow you to reproduce the shuffle order of a previous run
 	// to isolate an error condition.
-	Randomize randomSeed
+	Randomize int64
 
 	// Stops on the first failure
 	StopOnFailure bool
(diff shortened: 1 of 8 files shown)


mroth:
FWIW, I my simple method of testing this thus far was with a placeholder Feature file and `RunWithOptions` to enable randomization:
```
Feature: check execution order

  Scenario: Test number 4       # features/foo.feature:11
    Given this is a placeholder # godog_test.go:11 -> github.com/DATA-DOG/godog/examples/random.thisIsAPlaceholder
      TODO: write pending definition

  Scenario: Test number 2       # features/foo.feature:5
    Given this is a placeholder # godog_test.go:11 -> github.com/DATA-DOG/godog/examples/random.thisIsAPlaceholder
      TODO: write pending definition

  Scenario: Test number 3       # features/foo.feature:8
    Given this is a placeholder # godog_test.go:11 -> github.com/DATA-DOG/godog/examples/random.thisIsAPlaceholder
      TODO: write pending definition

  Scenario: Test number 1       # features/foo.feature:2
    Given this is a placeholder # godog_test.go:11 -> github.com/DATA-DOG/godog/examples/random.thisIsAPlaceholder
      TODO: write pending definition

  Scenario: Test number 5       # features/foo.feature:14
    Given this is a placeholder # godog_test.go:11 -> github.com/DATA-DOG/godog/examples/random.thisIsAPlaceholder
      TODO: write pending definition

5 scenarios (5 pending)
5 steps (5 pending)
```

options.go
```diff
@@ -13,6 +13,13 @@ type Options struct {
 	// Print step definitions found and exit
 	ShowStepDefinitions bool
 
+	// Run scenarios in random order.
+	//
+	// This is especially helpful for detecting situations
+	// where you have state leaking between scenarios, which
+	// can cause flickering or fragile tests.
+	RandomOrder bool
```
l3pp4rd: maybe could be `int` where `0` is not a random (should be explained in the comment and flag). Then this could be also available in flags. And you would not need to print it in suite, but instead if you run it through flags - you know it already, if from TestMain - you can print it there. Because godog might export junit xml to stdout if different format is used.

mroth:
#### Options and flags
Okay, I've moved to using an `int64` for propagating the option around, but in order to make the UI work in an non-confusing way for the user I ended up going with two different Option values -- check the commit history for seeing the evolution of my thinking here, but I believe the end result is more understandable from a UI perspective.  
```
Options:
  -c, --concurrency=1  Run the test suite with concurrency level:
                           - = 1: supports all types of formats.
                           - >= 2: only supports progress. Note, that
                           your context needs to support parallel execution.
  -d, --definitions    Print all available step definitions.
  -f, --format=pretty  How to format tests output. Available formats:
                           - cucumber: Produces cucumber JSON format output.
                           - events: Produces JSON event stream, based on spec: 0.1.0.
                           - junit: Prints junit compatible xml to stdout
                           - pretty: Prints every feature with runtime statuses.
                           - progress: Prints a character per step.
  --no-colors          Disable ansi colors.
  -o, --output=        Build and output test runner executable to given target path.
  --random             Randomize scenario execution order.
  --seed=SEED          Specify random seed to reproduce shuffling from a previous run.
                       (implies --random)
  --stop-on-failure    Stop processing on first failed scenario.
  -t, --tags=          Filter scenarios by tags. Expression can be:
                           - "@wip": run all scenarios with wip tag
                           - "~@wip": exclude all scenarios with wip tag
                           - "@wip && ~@new": run wip scenarios, but exclude new
                           - "@wip,@undone": run wip or undone scenarios
  --version            Show current version.
```

Check it out and let me know what you think!

(I also fixed a minor concurrency bug)

#### Output
Next step is I'd like to modify `basefmt.Summary()` to echo the random seed at the end in a matter similar to the ruby version of cucumber.  (For the record, that UI looks like this:)
```
10 scenarios (10 passed)
40 steps (40 passed)
0m3.220s

Randomized with seed 34128
```
I believe this should work fairly simply because all the formatters which display this sort of interactive output eventually call this, so we can get the behavior "for free" on appropriate formatters without negatively impacting ones designed for machine output (and of course, those could always be manually modified to insert the seed somewhere in their format if when there is an appropriate machine-readable field).

I'm having a bit of trouble figuring out an _elegant_ way to get the `Suite.randomSeed` value propagated from the struct to `basefmt` however, any thoughts around a good way to approach this?

options.go
```diff
@@ -13,12 +13,23 @@ type Options struct {
...
+	// Randomize causes scenarios to be run in random order.
 	//
-	// This is especially helpful for detecting situations
-	// where you have state leaking between scenarios, which
-	// can cause flickering or fragile tests.
-	RandomOrder bool
+	// Randomizing scenario order is especially helpful for detecting
+	// situations where you have state leaking between scenarios, which can
+	// cause flickering or fragile tests.
+	Randomize bool
```
l3pp4rd: this option becomes unnecessary, when seed is used it is inferred to be true. why not just changing it to one option **Randomize int64** which is a seed. default value `0` means do not randomize. A single option for single task, easier to understand from option list.

mroth: This is actually what I had in the first revised commit 0eed963 if you want to take a look. But you do need 3 categories of preference here:

1. do not randomize (most common case, with default nil value `0`),
2. randomize and let godog pick the seed (secondary common case),
3. randomize with a specific seed to try to reproduce a random failure (rare case). 

For the secondary case, I was originally solving in 0eed963 by having a magic value `-1` put as the seed.  This felt fine when I was interacting via Options struct.

However, once I later started mapping the Options struct to the command UI flags it became very confusing: passing something like `--random-seed=-1` for the more common randomization scenario was strange user interface.  Once I moved to the two variable model to have a more consistent mapping between the CLI and the Options struct, and it felt more understandable to work with.

It would also be possible to do something like this _just_ in the CLI and transform it based on rules before piping it back into the Options struct, but it appears that would require re-architecting the way Options/Flags interact currently.

For comparison, the way the CLI options on Ruby Cucumber CLI work are:
```
 --order random          (random)
 --order random:12345    (random with specified seed)
```

Let me know what you think!

l3pp4rd: Hm, indeed. that becomes confusing. Go allows to create custom flag mappings, maybe it would be possible to leverage those with something like:

1. do not randomize order (most common case, with default nil value `0`),
`godog` without any options.
2. randomize order and let godog pick a pseudorandom seed (secondary common case),
`godog --randomize` could set `GODOG_SEED` env variable in order to be able to check or print what seed was used.
3. randomize order with a specific seed to try to reproduce a random failure (rare case).
`godog --randomize $(GODOG_SEED)`

Need to try to experiment in order to check whether that would be possible. Otherwise, maybe the way cucumber ruby has implemented it, would be more suitable

l3pp4rd: I have made a little scenario which would work, see https://play.golang.org/p/CUW5cRjNtb that way we would have only one flag and no confusion.

mroth: @l3pp4rd that's very cool. I wasn't aware of the ability to have dynamic flags such as that in the default Go flag package. I'll try to take a stab at integrating this soon!

mroth:
(Weird, I typed a long version of this comment this morning but it seems to be gone now 😭)

Short version: your suggestion worked great!, check it out in the commit above.

Next up I want to figure out how to elegantly propagate the randomSeed value to `basefmt` so it can be displayed in `basefmt.Summary()`.  I don't yet see an obvious elegant way to do this without a lot of extra passing things around once it gets beyond the `Suite`. (One thought is to just use a package-scoped global variable so it can be referenced anywhere -- but I generally try to avoid global vars whenever possible).

l3pp4rd:
Hi @mroth I've made a [patch commit](https://github.com/DATA-DOG/godog/commit/5f894d572882295c52750350a3994ca6e180256a) on your branch clone. You can cherry pick it to your branch, so it could be in the pull request.

In your implementation `randomSeed` type was used in Options, which is not possible for the user, the type is not exported and we do not want to export it.

When running the suite from TestMain, user should generate the seed himself, that is to again, prevent exporting this seed type.

The seed will be stored in local environment in order to print it in summary.

mroth:
Looks great to me.  An additional benefit to storing it as a EnvVar is any subprocesses spawned by a scenario can access it if needed (can't think of a use case offhand, but still probably useful to someone!).

Cherry-picked the commit in.

l3pp4rd:
cool, accepting the change finally :) thanks

mroth:
Thank you! Fun working with you, I learned quite a bit.

PR #82, opened 2017-05-05
Reusable metatesting for add-on developers

### LINKED ISSUE #80: Request: move godog.SuiteContext into testing subpackage
In order to allow third-parties to develop their own libraries of reusable Godog steps, it should be easy for them to meta-test those steps themselves.

Currently, it looks like you have a great set of steps defined via `SuiteContext` in `suite_test.go`.  While it is technically "exported":

> Files matching `*_test.go` are only compiled when testing the package they are part of. If you are testing package A that uses package B, you won't have access to the` _test.go` code from package B.

(from http://stackoverflow.com/questions/22277537/export-structure-only-for-testing-in-golang)

I am trying to make a library of reusable godog steps, and it doesn't make sense to reinvent the wheel in testing them.  It would be fantastic to be able to use the SuiteContext.  Therefore I'd like to request you consider moving SuiteContext into its own subpackage in the godog hierarchy, and then import it in `suite_test.go`.  This would allow a third-party package to import this package during its own tests, so that it can also do meta-testing of features/steps.

mroth:
closes #80, this is the sample code for my proposal to make the `SuiteContext` itself general-purpose with an (optional!) variadic slice of additionalContextInitializers.

I believe this methodology will let a larger variety of developers make special-interest plugins to enhance the functionality of godog itself, and test them in a consistent fashion before offering to the public.

I personally would greatly prefer this general approach versus the of using external files and parsing jUnit XML proposed in https://github.com/DATA-DOG/godog/issues/80#issuecomment-299273733 for the following reasons:
1. No need to reinvent the wheel, potentially introducing bugs in the meta-testing framework itself. (As the existing SuiteContext is used for testing godog itself, it gives piece of mind to me as a add-on developer to know the same code is being used to test a potential godog add-on that it will work properly.
2. No need to write features to the filesystem enables easier concurrent-safe tests (due to filename collisions).
3. Having to develop a one-off test-harness to reliably replicate `SuiteContext` functionality externally feels discouraging to potential plugin developers, substantially increasing the scope of work needed to make a 
4. Easy re-use/discoverability by encouraging future add-on developers.

Cons:
1. End-user developers with no interest in developing godog addons may see `SuiteContext` in the godoc browser and it may increase the perceived complexity (I believe this could be potentially mitigated by renaming the function to something more obvious and writing documentation. I would be willing to help take this on if desired.)

That said, this is just my opinion. So thank you for your consideration! 🙇 

(This PR would at least need to be rebased off most-recent master before merging, but wanted to put it in a PR format to make it easier to review/discuss beforehand.)


suite_context.go
```diff
@@ -0,0 +1,486 @@
...
+	"github.com/DATA-DOG/godog/gherkin"
+)
+
+// SuiteContext can be used for meta-testing of godog features/steps themselves.
+//
+// A typical user of the godog library should never need this, rather it is
+// provided for those developing add-on libraries for godog.
+//
+// For an example of how to use, see godog's own `features/` and `suite_test.go`.
+func SuiteContext(s *Suite, additionalContextInitializers ...func(suite *Suite)) {
```
l3pp4rd: Lets change the doc a little

``` go
// SuiteContext provides steps for godog suite execution and
// can be used for meta-testing of godog features/steps themselves.
//
// Beware, steps or their definitions might change without backward
// compatibility guarantees. A typical user of the godog library should never
// need this, rather it is provided for those developing add-on libraries for godog.
//
// For an example of how to use, see godog's own `features/` and `suite_test.go`.
func SuiteContext(s *Suite, additionalContextInitializers ...func(suite *Suite)) {
```

suite_context.go
```diff
@@ -0,0 +1,486 @@
...
+	s.Step(`^this step should fail`, c.aFailingStep)
+	s.Step(`^the following steps? should be (passed|failed|skipped|undefined|pending):`, c.followingStepsShouldHave)
+	s.Step(`^all steps should (?:be|have|have been) (passed|failed|skipped|undefined|pending)$`, c.allStepsShouldHave)
+	s.Step(`^the undefined step snippets should be:$`, c.theUndefinedStepSnippetsShouldBe)
+
+	// event stream
+	s.Step(`^the following events should be fired:$`, c.thereShouldBeEventsFired)
+
+	// lt
+	s.Step(`^savybių aplankas "([^"]*)"$`, c.featurePath)
```
l3pp4rd: maybe some of those steps aren't necessary, probably can be extracted later for godog specific test cases. but lets leave it for now just for the sake of simplistic pull request

l3pp4rd:
please rebase and update the godoc, then we can merge. I'll later try to extract some of the steps, which aren't needed for developers, like ones related to cucumber json formatter tests or duplicates for multilingual features.

mroth:
Rebased but having some weirdness with the line numbers in the `cucumber.feature`. Interesting the line number matching differs when run with `-race -covermode=atomic` versus standard `go test`!  I updated to match the line numbers with that, which makes it pass on 1.8 in travis-ci, but not on 1.5-1.7, which has differing line numbers as well!  @l3pp4rd any idea what's going on here?

(Maybe I need more :coffee:)

suite_context.go
```diff
@@ -0,0 +1,488 @@
...
+		}
+		if err := s.thereWereNumEventsFired("", int(num), row.Cells[0].Value); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (s *suiteContext) theRenderJSONWillBe(docstring *gherkin.DocString) error {
+	loc := regexp.MustCompile(`"suite_test.go:\d+"`)
```
l3pp4rd: update to `suite_context.go`

l3pp4rd: when there is -race option, go adds more imports on the fly. and the line differs

mroth: updated to `suite_context.go`, weirdly enough looks like there is a difference between line numbers on 1.5-1.7 and 1.8.  Perhaps 1.8 changed the imports required?

mroth:
Okay, was missing one regex.  Fixed now.  Going to squash all the line number fun in a second.

mroth:
Squashed! 🐛

l3pp4rd:
cool, thanks

PR #83, opened 2017-05-05
Error details from various state

l3pp4rd:
This adds an ability to append context to an error, from various feature contexts.
Sometimes it is important to have some state listed when a step fails, this state may be gathered from a number of different contexts.

This example shows how it can be used. What is your opinion on this @mroth in general, user might create and wrap error in the same way, but that would introduce some boilerplate. Though, godog may manage it correctly for various formatters and indentation.


mroth:
Hmm, I can't immediately think of a case where I don't have access to the context in my actual steps (since I'm typically writing them as methods on a context), but since this only gets wrapped later, from an end-user perspective you can still simply return a normal `error` when writing a step, it doesn't seem to add too much complexity.

It may also be possible to do this by simply rewriting the error message (e.g. appending `"\n" + foo`) in the after step, but that obviously gives you less formatting control than a collection of attachments.

l3pp4rd:
For big projects, you may have various contexts. Like: QueueContext, APIContext, SeleniumContext, UserContext all of them may provide some steps and some of them may have internal state, which may be useful to see if one of steps fails, like

- SeleniumContext - making a screenshot and saving html body, if selenium is used for that scenario
- QueueContext - may dump all items currently in the queue
- APIContext - could dump last response received information
- UserContext - currently logged in user, his permissions

In general, each of these contexts, could append some of the details after any **step** fails. That often makes things much more visible and saves debugging time.

Since this extension, does not add much to public interface and will most likely be useful for attachment like behavior, which may be built in into the gherkin, I may include it later after some more testing.

examples/api/version.feature
```diff
@@ -10,7 +10,7 @@ Feature: get version
     And the response should match json:
       """
       {
-        "error": "Method not allowed"
+        "error": "ethod not allowed"
```
YoussefKaib: fix typo `method`

l3pp4rd: yes thanks, but most likely this pull request won't be merged, as all hooks will be refactored at some point to return an error, to follow the idiomatic go way, see #88 that way it will be possible to simple extend error message in a hook.

YoussefKaib: Glad to hear! As a STE automating acceptance criteria getting meaningful message is important for debugging. currently it's hard with current dodog version.

lonnblad:
@l3pp4rd should we close this PR?

l3pp4rd:
Yes should be closed and instead hooks should return error

PR #109, opened 2017-12-08
Add flag to enable writing results to a file

darwish:
This change adds a flag that enables writing results directly to a file, rather than needing to pipe the output.

The motivation for this is that we have a large series of tests that print debugging information to stdout, and this output gets mixed in with the result JSON, making it unparseable.

Changing the `opt.Output` property doesn't work in our case, because we use the `--output` flag to generate a binary and then run the binary in a separate environment.


l3pp4rd:
hi @darwish sadly I cannot accept it because it would force to cause a BC break in the future. original cucumber implementation in **ruby** allows to specify various outputs with different formatters like `cucumber --format=progress@output/progress.out --format=junit@junit.xml --format=pretty@stdout`. And I'm thinking to implement something like this in the future, just not yet decided how.

So as a solution I suggest you to compile test executable which runs godog by overriding TestMain, you can compile tests executable the same way as binary `go test -c -o tests`

darwish:
@l3pp4rd No worries, thanks for the pointer.

PR #116, opened 2018-03-02
Fixed: compilation of godog under golang 1.10

Dogild:


Dogild:
@l3pp4rd If this seems good/ok for you let me know. I'll add the unit tests for the function `isVersionGreaterOrEqual`

l3pp4rd:
Hi @Dogild I tried exactly these changes so far in [branch](https://github.com/DATA-DOG/godog/tree/go1.10) but there is an issue with cache, try to clean it `rm -rf ~/.cache/go-build`

l3pp4rd:
ah while using the `importcfg.link` it works, but only with go files available. in previous versions godog was able to run in non `GOPATH` directory or without any go files at all, only a feature file. that is still missing for compatibility

l3pp4rd:
it is more complex than that. thanks for the effort, you gave me a good hint regarding the importcfg.link. was helpful. I have opened a pull request, only vendor package dependency support is missing still

Dogild:
Yep, it was a first step :) thanks for the rest of the work!

PR #123, opened 2018-03-30
Fix windows /dev/null bug with go1.10.0

Pykselle:
Since last update on go 1.10.0 and godog v0.7.7, a "dev" folder (containing the test binary named "null") was created in the current test directory (where the "godog" cmd was called) on WINDOWS machines


l3pp4rd:
ah, good. thanks ;)

PR #138, opened 2018-09-29
godog command is on error with go modules and sub-packages

rvflash:
With go 1.11 and the environment variable GO111MODULE enabled, the godog command is on error and fails to compile the `testmain` because the path used as import path is not the good one (only the module path is used).

Example of result when godog is called from the directory `module/qa`: 

`failed to compile testmain package: exit status 2 - output: /var/folders/6c/k20xl1c17735407tz10y6xs80000gn/T/go-build466142923/b001/_testmain.go:5:2: can't find import: "module"`

I hope it helps, I have also fixed some lints.

### DIFF
--- a/builder.go
+++ b/builder.go
@@ -241,6 +241,8 @@ func makeImportValid(r rune) rune {
 	return r
 }
 
+type void struct{}
+
 func uniqStringList(strs []string) (unique []string) {
 	uniq := make(map[string]void, len(strs))
 	for _, s := range strs {
--- a/builder_go110.go
+++ b/builder_go110.go
@@ -19,16 +19,15 @@ import (
 	"unicode"
 )
 
-var tooldir = findToolDir()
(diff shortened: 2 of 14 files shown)


l3pp4rd:
Hi @rvflash , thanks, I will check this properly in few days

rvflash:
Hi @l3pp4rd, thanks.

I had to make a new modification after a last test with a `go.mod` file inside the GOPATH.
Not sure this is the best way to do it, the go/build package lacks means to deal with modules.

At your disposal if needed,

l3pp4rd:
merging it for now, later will need to create proper build tests, but do not worry about it

l3pp4rd:
could you also give an example of this module subpackage build failure? is it a case when you have lets say a module at `/tmp/module` and a subpackage in `/tmp/module/sub` which was not building properly before?

PR #141, opened 2018-10-03
implementation to convert datatable to struct

matheussilva97:
useful way to convert data table of feature to struct in a simple and fast way.

Example:

Data Table:

```
  | value 1  | value 2 | result |
  |  5         |  10     |   15       |
```
Struct:

```
type Table struct {
	Value1 string `godog:"value 1"`
	Value2 string `godog:"value 2"`
	Result string `godog:"result"`
}
```
usually:

```
 var t Table 
 gherkin.ParseDataTable(dataTable, &t)
 fmt.Println(t.Value1) // print 5
 fmt.Println(t.Value2) //print 10
 fmt.Println(t.Result)  //print 15
```

### DIFF
--- a/gherkin/gherkin.go
+++ b/gherkin/gherkin.go
@@ -4,6 +4,8 @@ import (
 	"bufio"
 	"fmt"
 	"io"
+	"reflect"
+	"strconv"
 	"strings"
 )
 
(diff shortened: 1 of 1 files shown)


l3pp4rd:
@matheussilva97 in [my comment](https://github.com/DATA-DOG/godog/issues/140#issuecomment-426747017) I explained why godog will not maintain these kind of features. my suggestion was to make that implementation in your own test files.

matheussilva97:
Ok, Sorry about that, I understand wrong. I thought that this functionality could be useful for other people, but ok, I'm sorry @l3pp4rd

l3pp4rd:
it could be useful, but not in this main repository, users have various needs. I cannot add such utilities, they cannot be maintained because it will never make everyone happy. you can create library in your own repository, people may find and use it.

matheussilva97:
Ok, thanks for the feedback :D

PR #144, opened 2018-10-25
Allow overriding flag defaults in BindFlags

### LINKED ISSUE #143: Update `BindFlags` to not overwrite existing `Options` with default values
If my test runner has code similar to:

```
var godogOpts = godog.Options{
  Format:    "progress",
}

func init() {
  godog.BindFlags("godog.", flag.CommandLine, &godogOpts)
}

func TestMain(m *testing.M) {
  flag.Parse()
  ...
}
```

the current implementation of [`BindFlags`](https://github.com/DATA-DOG/godog/blob/master/flags.go#L47-L67) overwrites my supplied `Format` with the default `"pretty"`, even if I don't specify a `godog.format` flag. 

I would think there's a valid use case for wanting to codify default values, but allow overrides to those values as command line flags. Is there an existing way to do this? Would an update to support this functionality in `BindFlags` be welcome?

patrickvalle:
This is an implementation of the feature request @ https://github.com/DATA-DOG/godog/issues/143. `BindFlags` has been updated to use any non-empty `opt` properties as default overrides. For any empty `opt` properties, the default... defaults are used.

closes #143

### DIFF
--- a/flags.go
+++ b/flags.go
@@ -52,17 +52,47 @@ func BindFlags(prefix string, set *flag.FlagSet, opt *Options) {
 	}
 	descFormatOption = strings.TrimSpace(descFormatOption)
 
-	set.StringVar(&opt.Format, prefix+"format", "pretty", descFormatOption)
-	set.StringVar(&opt.Format, prefix+"f", "pretty", descFormatOption)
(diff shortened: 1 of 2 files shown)


l3pp4rd:
nice work, thanks ;)

PR #168, opened 2019-03-26
Make godog.Suite.RunSteps a public method

### LINKED ISSUE #167: Make `Suite.runSteps` Method public
I'm using godog in a complex system testing scenario and would really like to more flexibly reuse parts of my feature files. I have been using the multistep feature but think it could be replaced by this functionality as it reuses the Suite to run gherkin scenarios which can be parsed from external files or `gherkin.DocString` without the restrictions on `Datatable`s or `DocString`s as arguments metioned here #89 

```go
func FeatureContext(s *godog.Suite) {
	s.Step(`^the "([^"]*)" passes$`, makeTheFeaturePasses(s))
}

func makeTheFeaturePasses(context *godeliatypes.Context) func(string) error {
	return func(featureFile string) error {
		file, err := os.Open(featureFile);
		if err != nil {
			return fmt.Errorf("couldn't open feature file: %v", err)
		}

                feature, err := gherkin.ParseFeature(file)
		if err != nil {
			return fmt.Errorf("couldn't parse feature file: %v", err)
		}

		for i := range feature.ScenarioDefinitions {
                        if d, ok := feature.ScenarioDefinitions[i].(*gherkin.Scenario); ok { 
				return context.Suite.RunSteps(d.Steps)
			} else { 
                                return fmt.Errorf("couldn't handle %v", feature.ScenarioDefinitions[i])
                        }
		}

		return nil
	}
}

```

```gherkin
Feature: Another Multi Step Feature

    Scenario: Run the external Feature
        Given the "reusable_backgrounds/preparation.feature" passes
```

ziprandom:
Make `godog.Suite#runSteps` public `godog.Suite#RunSteps`. 

Reuse a already running Suite to run custom parsed gherkin steps against it. 

closes #167

### DIFF
--- a/suite.go
+++ b/suite.go
@@ -383,7 +383,7 @@ func (s *Suite) matchStepText(text string) *StepDef {
 	return nil
 }
 
-func (s *Suite) runSteps(steps []*gherkin.Step) (err error) {
+func (s *Suite) RunSteps(steps []*gherkin.Step) (err error) {
 	for _, step := range steps {
 		stepErr := s.runStep(step, err)
 		switch stepErr {
@@ -487,7 +487,7 @@ func (s *Suite) runOutline(outline *gherkin.ScenarioOutline, b *gherkin.Backgrou
 				steps = append(b.Steps, steps...)
 			}
 
-			err := s.runSteps(steps)
+			err := s.RunSteps(steps)
 
 			if !isEmptyScenario(outline) {
 				for _, f := range s.afterScenarioHandlers {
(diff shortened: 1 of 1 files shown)


suite.go
```diff
@@ -383,7 +383,7 @@ func (s *Suite) matchStepText(text string) *StepDef {
 	return nil
 }
 
-func (s *Suite) runSteps(steps []*gherkin.Step) (err error) {
+func (s *Suite) RunSteps(steps []*gherkin.Step) (err error) {
```
bkielbasa: how about creating the public function which calls the internal one?
IMO you're missing the docs as well :)

jaysonesmith:
[Please see my comment here for context.](https://github.com/cucumber/godog/issues/167#issuecomment-830236656)

PR #171, opened 2019-04-14
Use reflect.DeepEqual to compare json strings

joeky888:
Source here https://gist.github.com/turtlemonvh/e4f7404e28387fadb8ad275a99596f67


l3pp4rd:
hm, that looks good, thanks @joeky888 ;)

PR #174, opened 2019-05-21
Added Support for Cucumber JSON Report's File Embeddings

cbush06:
To support embedding screenshots and other files into the Cucumber JSON Report Format, I added necessary features to GoDog that allows this.

I've also updated the documentation to include a section describing this feature under the "FAQ" heading.

For an example of this feature in use, see these 2 files:

* https://github.com/cbush06/kosher/blob/64f98bceae8028963353ddd6ea4538a61330aece/steps/websteps/browser_steps.go#L34
* https://github.com/cbush06/kosher/blob/64f98bceae8028963353ddd6ea4538a61330aece/steps/websteps/web_steps.go#L20

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,5 @@
 /cmd/godog/godog
 /example/example
+**/vendor/*
+Gopkg.lock
+Gopkg.toml
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ go:
   - 1.9.x
   - 1.10.x
   - 1.11.x
+  - 1.12.x
 
 go_import_path: github.com/DATA-DOG/godog
 
--- a/README.md
+++ b/README.md
@@ -358,6 +358,55 @@ it down only in **AfterFeature** hook and share the service between all
 scenarios in that feature. It is not advisable though, because you are
 risking having a state dependency.
 
+### Support for Cucumber Embeddings
+
+**godog** has support for Cucumber's JSON Report file embeddings. To add an
+embedding during a step, add one final parameter to the step function.
+*The step definition's regex captures do not need to include a capture that*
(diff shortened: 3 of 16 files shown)


l3pp4rd:
Hi, we cannot modify the gherkin library sources, at least not for now, since [this library](https://github.com/cucumber/gherkin-go) is extracted as a subpackage from cucumber repository. that would be very painful to update it if we have the customizations in our end.

The reasonable way to implement this as I see it, would be to add `godog.Suite.Embed(loc *gherkin.Location, mimeType string, data []byte)` method. which would directly use a `sync.Map` in Suite.fmt and embed it there. Based on that **gherkin.Location** we would always find a correct gherkin step, or **gherkin.TableRow** which took the screenshot and embedded it. Then the formatter may extract the embedding from that map.

This would also allow to add hooks, so that after every failing or passing step a screenshot can be made and embedded.

cbush06:
@l3pp4rd I see what you mean. I'll refactor my solution following your advice.

inluxc:
Have you guys notice the new update on Gherkin Repo...
This looks interesting for this PULL

https://github.com/cucumber/cucumber/blob/master/cucumber-messages/go/messages.pb.go#L2093

jaysonesmith:
@cbush06 Thanks for all your help here. Let's hold off on this for now until we've determined the direction the tool will take and what's missing/required then. Going to close this for now, but feel free to reopen it in the future if it still fits. :)

andrewesweet:
Is there any interest in re-opening this? I’m interested in being able to embed plain text and _may_ have some time in the not-too-distant future to invest in it.

Johnlon:
It's 2024 now - any news on enabling Embeddings as per Cucumber?

Johnlon:
Perhaps a fork?

PR #176, opened 2019-05-29
fix(output): use bold-black for muted texts

### LINKED ISSUE #121: adjust pretty print test result color to be more visible
the current color for test result comments is in black color, this makes it really hard to see when the termial is in black background.

Can we please change it to a color that suitable for both dark and light background?

screen shot 2018-03-21 at 6 04 54 pm
screen shot 2018-03-21 at 6 05 30 pm

kumarharsh:
Closes #121


l3pp4rd:
thanks ;)

Issue #161, opened 2019-02-15
Summary with Concurrency has race condition


smikulcik:
The summary output has incorrect Scenario Step, and Line number references when running with concurrency. 

## Actual Output
```
go test -v . --godog.format=progress --godog.concurrency=2 ./features/
........F-......F-F- 20


--- Failed steps:

  Scenario: RUN2 I do a bunch of pass but fail # features/run1.feature:10
    Then I fail RUN1 # features/run1.feature:7
      Error: I fail

  Scenario: RUN1 I do a bunch of pass but fail # features/run1.feature:16
    Then I fail RUN2 # features/run1.feature:13
      Error: I fail

  Scenario: RUN1 I do a bunch of pass but fail # features/run1.feature:16
    Then I fail RUN1 # features/run1.feature:19
      Error: I fail


5 scenarios (2 passed, 3 failed)
20 steps (14 passed, 3 failed, 3 skipped)
364.968µs
testing: warning: no tests to run
PASS
FAIL    godogsummaryfail        0.045s
```

## Expected Output
```
go test -v . --godog.format=progress --godog.concurrency=2 ./features/
........F-......F-F- 20


--- Failed steps:

  Scenario: RUN1 I do a bunch of pass but fail # features/run1.feature:4
    Then I fail RUN1 # features/run1.feature:7
      Error: I fail

  Scenario: RUN2 I do a bunch of pass but fail # features/run2.feature:10
    Then I fail RUN2 # features/run1.feature:13
      Error: I fail

  Scenario: RUN1 I do a bunch of pass but fail # features/run1.feature:16
    Then I fail RUN1 # features/run1.feature:19
      Error: I fail


5 scenarios (2 passed, 3 failed)
20 steps (14 passed, 3 failed, 3 skipped)
364.968µs
testing: warning: no tests to run
PASS
FAIL    godogsummaryfail        0.045s
```

## 

 ## Setup

using github.com/DATA-DOG/godog 0.7.10

steps_test.go
```go
// file: steps_test.go
package main

import (
	"flag"
	"fmt"
	"os"
	"testing"

	"github.com/DATA-DOG/godog"
	"github.com/DATA-DOG/godog/colors"
)

var opt = godog.Options{Output: colors.Colored(os.Stdout)}

func init() {
	godog.BindFlags("godog.", flag.CommandLine, &opt)
}

func TestMain(m *testing.M) {
	flag.Parse()
	opt.Paths = flag.Args()

[...]

PR #190, opened 2019-08-31
fix for progress formatter summary in concurrency mode

funvit:
fix for #161

I cant provide test(s), only checked manually...

Bug fixed via creating custom formatter scope for each suite.run() and coping results to runner formatter.


funvit:
Hmm. My fix breaks progress output for steps, so i need to fix that....

PR #191, opened 2019-09-02
fix for progress formatter summary in concurrency mode

funvit:
fix for #161

Bug fixed via creating custom formatter scope for each suite.run() and coping results to runner formatter.

1 test added.


l3pp4rd:
thanks, merging it since do not have time to properly revisit the gherkin to prevent dealing with state

PR #201, opened 2019-12-15
Fixed examples directory

arberiii:
To run godog in examples/api or examples/db, feature files should be in their directory.


l3pp4rd:
thanks ;)

PR #206, opened 2020-01-20
Send report to Report Portal

inluxc:
This is an WIP for sending reports to Report Portal.

https://github.com/reportportal/reportportal


l3pp4rd:
this is very specific to certain use case. can you rather implement it as your own separate package.

jaysonesmith:
Hey @inluxc! Thanks for the work in rigging this up, but I don't think that it fits in the context of the tool itself. Similar to what @l3pp4rd said, implementing it as your own package should be the route to go. Thanks!

PR #220, opened 2020-02-10
Added gherkin verification of the pretty formatter

lonnblad:


lonnblad:
We should probably remove the fixtures I added for JUnit and Cucumber and focus on more `features/formatter/<formatter>.feature`

jaysonesmith:
@lonnblad Why do you say we should remove them? Are they not providing any value currently?

lonnblad:
I believe they provide som value, but we can run them with less feature files, just to verify that the concurrency works and with that at least have much smaller fixtures.

The main verification of the formatters would be done like we already do in `features/formatter/cucumber.feature` and what I know did with the `features/formatter/junit.feature`.

jaysonesmith:
That sounds good. I'll trust your judgement!

Issue #222, opened 2020-02-11
[Enhancement] Allow access to GherkinDocument.GherkinDocument and messages.Pickles from godog.Suite


radtriste:
In our tests, before the suite is executed, we need to checks the features for specific tags.
The problem is that there is no way to get those from `godog.Suite`.

I understand you may not want to disclose the internal feature structure but it would be good to have a `getFeatures` method that would allow to get all gherkin features directly from the suite.

lonnblad:
Since we are in the development of migrating to https://github.com/cucumber/gherkin-go, v9.2.0.
I think we should wait with this one, because it will most probably get deprecated as soon as we are done with the migration.

coltiebaby:
I believe you can @radtriste.

If you set up a `BeforeFeature` inside your context, you should be able to get what you need.
We do something along the lines of this to check a workdoc to see if a test needs to be skipped.

```
func SetupContext(c *CommonContext) func(*godog.Suite) {
    return func(s *godog.Suite) {
         s.BeforeFeature(func(feature *gherkin.Feature) {
                // Do something with feature or scenario outlines.
         } )
    }
}
```

Because we use scenario outlines, we can easily delete a row if we do not want to run it. Not sure what your tags will do but if they modify some value this should do it too.

radtriste:
I thought about that but in fact, depending on the tags used, the smoke tests may need or not an external binary. And I want to check that this binary is available (if I have some filtered features that need it) before the tests are launched. So I need to have the list of features/scenarios which will be executed.
I could just panic in before feature or before scenario if binary is not available but I don't like it much as it breaks all other checks and is not reported as a simple scenario error (no panic catch in those Before* methods).

lonnblad:
Hi @radtriste, would it be enough if you get them in a `BeforeScenario` hook?
I will get both the feature and scenario tags there.

radtriste:
Hello @lonnblad 
My problem is that I want to be able to decide in the main part if tests can be run or not.
If the binary I mention in https://github.com/cucumber/godog/issues/222#issuecomment-584697747 is not present, then all the tests should not be executed and an error is reported.
The `BeforeScenario` will no be enough. 

I need to be able to make a decision based on all scenario which ill be executed but before the start of execution of those.
And for now I am not able to do so.
That is why https://github.com/cucumber/godog/pull/276 is useful.

lonnblad:
Would you supply that tag when you execute the tests?
like: ```godog --tags "smoke-test"```

[...]

PR #223, opened 2020-02-11
Added GetFeatures method to godog.Suite

radtriste:
https://github.com/cucumber/godog/issues/222


radtriste:
clsing and waiting for https://github.com/cucumber/godog/pull/227 to be merged

PR #224, opened 2020-02-11
Added gherkin verification of the JUnit formatter

lonnblad:


lonnblad:
we should fix so that we can verify test-case durations in some way as well, when I did the concurrent JUnit formatter, it would really have helped me

PR #240, opened 2020-02-29
Migrated to github.com/cucumber/gherkin-go - v9.2.0

lonnblad:


fmt_cucumber.go
```diff
@@ -122,115 +121,86 @@ type cukefmt struct {
 	// of the example name inorder to build id fields.
 }
 
-func (f *cukefmt) Node(n interface{}) {
-	f.basefmt.Node(n)
+func (f *cukefmt) Pickle(pickle *messages.Pickle) {
+	f.basefmt.Pickle(pickle)
 
-	switch t := n.(type) {
+	scenario := f.findScenario(pickle.AstNodeIds[0])
```
leviable: Should there be a `if scenario == nil` check here? Otherwise wouldn't line 133 panic?

lonnblad: I'm not sure, if `f.findScenario` doesn't find a scenario, something is really broken, but we could add something like this:
```go
if scenario == nil {
  panic("couldn't find scenario for AST ID: " + astNodeID)
}
```

to improve debugging

suite.go
```diff
@@ -724,39 +670,36 @@ func parseFeatureDir(dir string) ([]*feature, error) {
...
-	path, line = extractFeaturePathLine(path)
+
+	path, line := extractFeaturePathLine(path)
 
 	fi, err := os.Stat(path)
 	if err != nil {
 		return features, err
 	}
 
+	newIdFunc := (&messages.Incrementing{}).NewId
```
jaysonesmith: Go should yell at you for this variable name. :p It should be newIDFunc instead of newIdFunc

lonnblad: haha, true, will fix

would be nice to add more linting in general

lonnblad: I don't think our linting is running as it should btw. it misses more things, like the colors pkg.

fmt_cucumber.go
lonnblad: what do you think about this:
```
func (f *basefmt) findScenario(scenarioAstID string) *messages.GherkinDocument_Feature_Scenario {
	for _, ft := range f.features {
		if sc := ft.findScenario(scenarioAstID); sc != nil {
			return sc
		}
	}

	panic("couldn't find scenario for AST ID: " + scenarioAstID)
}
```

suite.go
jaysonesmith: Can look into that. I personally use the go extension in VS Code which lints and shows issues on its own. It’s awesome! :p

fmt_cucumber.go
leviable: 👍

PR #272, opened 2020-03-24
Added comparison between single and multi threaded runs

lonnblad:


run.go
```diff
@@ -274,7 +276,11 @@ func supportsConcurrency(format string) bool {
 	switch format {
 	case "progress", "junit":
 		return true
-	case "events", "pretty", "cucumber":
+	case "events":
```
jaysonesmith: Was this not working before?

lonnblad: it was, but I made this change to have easier merges later,`supportsConcurrency` will be removed as soon as all formatters updates are merged

jaysonesmith: Works for me :)

Issue #286, opened 2020-04-29
Commenting out features causes nil pointer dereference


jamescooke:
Please answer these questions before submitting a bug report.

### What version of godog are you using?
```
> godog -version
Godog version is: v0.9.0
```
Fails with both `go get` of `v0.9.0` and `master`.

### What version of Go are you using?
```
> go version
go version go1.14.2 linux/amd64
```

### What did you do?
Comment out all features in a feature file:

```cucumber
# Feature: list Items
# 
#   API clients can do `GET /items` to retrieve a list of Items.
# 
#   Scenario: Items are empty
#     When API client requests list of Items
#     Then response status is 200
#     And response contains an empty list of Items
```

### What did you expect to see?

Either a warning that no feature was found, or silently skip. Godog should execute the other features in other feature files that are not commented out.

### What did you see instead?

Run `godog` and get a stack trace:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x38 pc=0x8f0fff]

goroutine 1 [running]:
github.com/cucumber/godog.(*pretty).printFeature(0xc000010268, 0x0)
        /home/vagrant/go/pkg/mod/github.com/cucumber/godog@v0.9.0/fmt_pretty.go:75 +0x4f
github.com/cucumber/godog.(*pretty).Feature(0xc000010268, 0xc00012d740, 0xc00002a780, 0x1c, 0xc00014a000, 0x178, 0x178)
        /home/vagrant/go/pkg/mod/github.com/cucumber/godog@v0.9.0/fmt_pretty.go:32 +0x8a
github.com/cucumber/godog.(*Suite).runFeature(0xc00014f680, 0xc0000c6880)
        /home/vagrant/go/pkg/mod/github.com/cucumber/godog@v0.9.0/suite.go:543 +0xca
github.com/cucumber/godog.(*Suite).run(0xc00014f680)
        /home/vagrant/go/pkg/mod/github.com/cucumber/godog@v0.9.0/suite.go:339 +0xaf
github.com/cucumber/godog.(*runner).run(0xc000119f80, 0x14)
        /home/vagrant/go/pkg/mod/github.com/cucumber/godog@v0.9.0/run.go:118 +0xae
github.com/cucumber/godog.RunWithOptions(0x9ff1c6, 0x5, 0xc00010d740, 0x0, 0x0, 0x0, 0x0, 0x0, 0x9ffa17, 0x6, ...)
[...]

PR #288, opened 2020-04-30
Fixed issue with empty feature file causing nil pointer deref

lonnblad:
Fix for #286


jamescooke:
Thanks for the quick fix - this looks like it fixes #286 for me locally :+1:

PR #359, opened 2020-11-24
Fixed an issue when go test is parsing command-line flags

lonnblad:


lonnblad:
We should try and add a test case that covers this scenario.

titouanfreville (approved):
I like this solution more than adding a function that will do the trick. Think its clearer like this, less magical :)

PR #364, opened 2020-12-10
feat(step_definition): Allows to define step function without return.

titouanfreville:
Issue: It is not possible to use function without return when
       matching steps, resulting in a lot of Nil only error
       returns.

Fix: Allows to provide empty result function by correctly matching
     reflect Calls on step Handler.
     When nothing is returned by the Handler, it will return
     nil as if errors was nil.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ This document is formatted according to the principles of [Keep A CHANGELOG](htt
 
 ### Added
 
+* Support for step definitions without return ([364](https://github.com/cucumber/godog/pull/364) -[titouanfreville])
+
 ### Changed
 
 * Upgraded gherkin-go to v19 ([402](https://github.com/cucumber/godog/pull/402) - [mbow])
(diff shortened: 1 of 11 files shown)


titouanfreville:
Last commit add `goconvey` dependencies to simplify some unit test writting on Panics. 

It can be rollbacked or adapt to keep the tests with a custom system if dependencies should not be added.

mattwynne:
This needs re-basing before it can be merged. @vearutop could you help?

vearutop (approved):

vearutop:
Let's wait for @titouanfreville response for couple more days, otherwise this PR looks good to me to merge.

mattwynne:
I suggest updating the CHANGELOG as part of the PR, it's a good idea so you don't forget to do it after the merge.

titouanfreville:
If there are no conflict, it should be ok to merge it. 

I also used it to start a discution arround test tools but did not get any feedback on it.

I'll update the changelog ASAP before merging :)

vearutop:
I'm not opposed to new testing tools (since it does not affect end-user much) if they help writing more reliable and readable tests, though not sure if goconvey brings better ergonomics compared to `testify` or `testing` that are already in use.

vearutop:
Seems @titouanfreville is busy, so I updated README and CHANGELOG and going to merge this PR. If any additional changes are necessary, let's deliver them in a separate PR.

Issue #328, opened 2020-07-03
Request: option to provide features not from filesystem


zhammer:
Hi, I'm curious if y'all would be up allowing a way to provide feature files to `godog.TestSuite` that is not just passing file paths. Our use case is that we'd like to package a test tool as a binary using godog and ship features as static files in the binary using https://github.com/markbates/pkger. Here we'd have he feature file bodies and names; however, the TestSuite wouldn't be able to parse them from a list of filesystem Paths.

Happy to discuss different ways this could work!

https://github.com/cucumber/godog/blob/28ad994ad150cf84838309251c8f23a46fee7b35/options.go#L54-L55

lonnblad:
Sounds interesting, maybe when building the godog testrunner we could have an include functionality that would embed the feature files you pass.

Something like: `> godog build -i -o godog.test features` would build a test runner binary which embeds the features dir.

zhammer:
Yep that’d be nice. We have our own golang cli that uses the godog package directly so wouldn’t immediately need the cli option, but I do like the idea of including packaged feature files in the godog cli tool itself

lonnblad:
Were you thinking something like:
```
type GherkinFile struct {
  Name    string
  Content []byte
}

type Options struct {
  ...
  Files []GherkinFile
}
```

and let your code read the content from pkger?

lcarrazana:
Looking for the exact same functionality.
I'm using pkger for embedding all static files, and would like to also do it for godog feature files.

lonnblad:
@lcarrazana would you want to use it in a similar way as zhammer, or would you use the godog CLI directly?

vearutop:
Done with https://github.com/cucumber/godog/pull/476.

PR #374, opened 2021-02-09
Added public logic for pre-parsing features

eddie-knight:
This PR is to resolve issue #328.

Our project uses Godog to package a series of tests into a single binary that our users can execute in their own pipelines to validate their cloud infrastructure security compliance. 

Currently we have been using Pkger to store the feature files into the binary, write the feature files to a temporary directory that Godog can read, and then delete the files at the end of the runtime. This is functional, but not desirable.

The changes proposed in this PR will allow Godog users (such as our project) to pre-parse their feature files, instead of relying on I/O.

The following changes are present:
1. Expose the parser for public usage
2. Extend `flags.Options` to receive parsed feature objects
3. Add a condition to `run.go` that will use the parsed features instead of `Paths`
4. Extended existing tests to ensure "Features" are run identically to "Paths"

### DIFF
--- a/internal/flags/options.go
+++ b/internal/flags/options.go
@@ -2,6 +2,8 @@ package flags
 
 import (
 	"io"
+
+	"github.com/cucumber/godog/internal/models"
 )
 
 // Options are suite run options
@@ -51,8 +53,11 @@ type Options struct {
 	// Concurrency rate, not all formatters accepts this
 	Concurrency int
 
-	// All feature file paths
+	// All feature file paths. Cannot be used with Options.Features
 	Paths []string
+
+	// Parsed feature objects. Overrides Options.Paths
+	Features []*models.Feature
 
 	// Where it should print formatter output
 	Output io.Writer
--- a/parser.go
+++ b/parser.go
@@ -1,0 +1,14 @@
+package godog
+
+import (
+	"github.com/cucumber/godog/internal/models"
+	"github.com/cucumber/godog/internal/parser"
+)
+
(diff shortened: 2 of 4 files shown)


lonnblad:
Hi @eddie-knight, thanks for the PR!

Have you considered sending in the raw Content and URIs of the files instead of doing pre-parsing?
And having the Options to be something like:
```go
type Options struct {
	// ...

	// All feature file paths to be searched and parsed.
	// Overriden by Options.FeatureFiles.
	Paths []string

	// All feature file contents to be parsed.
	// Overrides Options.Paths.
	FeatureFiles []struct {
		URI string
		Content []byte
	}

	// ...
}
```

BR Fredrik

eddie-knight:
Thanks for the reply @lonnblad 

> Have you considered sending in the raw Content and URIs of the files instead of doing pre-parsing?

If we're passing in raw file content (pre-parsed) that would solve the problem for sure. Can you help me understand the need for the URI value when passing the file content? If I were to create another solution following your proposal, what would that URI value be used for?

> If the idea is that it's up to godog to search for and filter the feature files

Actually the idea proposed here is to give the option to for the user to either handle the files themselves _or_ leave it up to godog to search for and filter the files.

Is there anything specific in the proposed solution that isn't acceptable? Perhaps if I understand what needs changed then I'll better understand the alternative you're proposing.

eddie-knight:
Closing this as it seems like it isn't needed for the project, and my team has adjusted to solve the aforementioned issue in another way

PR #391, opened 2021-05-30
remove renamed imports of pflag.

rickardenglund:
This was causing way to much investigating in order for me to find
That I've only copied parts of the examples resulting in "flag".Parse
instead of pflag.Parse()

### DIFF
--- a/README.md
+++ b/README.md
@@ -346,13 +346,12 @@ The following example binds **godog** flags with specified prefix `godog` in ord
 package main
 
 import (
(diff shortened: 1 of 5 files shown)


aslakhellesoy:
This looks good to me, and if you think it's ready to merge I'll do it @rickardenglund.

We're discussing a merge of godog/gobdd over in the #committers-go channel in the [Cucumber slack](https://cucumberbdd-slack-invite.herokuapp.com/) - any chance you could join the discussion there?

PR #402, opened 2021-06-28
update gherkin-go to v19.0.3

mbow:
update messages-go to v16.0.1
bump gomod version
comment on log line in std os.Stderr
examples to non rc version
go mod tidy
update circle (tbd)

# Description

fixes for https://github.com/cucumber/godog/pull/390
this branch is to fix panics on an update of library versions due to getter methods being removed and default values are nil.

	"github.com/cucumber/gherkin-go/v19"
	"github.com/cucumber/messages-go/v16"

If it fixes an open issue https://github.com/cucumber/godog/issues/396

this should close a CVE in the above lib and allow use in orgs that scan for CVE's

## Type of change

- Refactoring

# Checklist:

- [X] My code follows the code style of this project.
- [-] My change requires a change to the documentation.
- [-] I have updated the documentation accordingly.
- [-] I have read the [**CONTRIBUTING**](../CONTRIBUTING.md) document.
- [-] I have added tests to cover my changes.
- [X] All new and existing tests passed.

Issue #389, opened 2021-05-21
Update cucumber/common module references to use paths instead of subrepos


mattwynne:
We have an initiative (https://github.com/cucumber/common/issues/1550) to remove some technical debt in the cucumber/common repo around publishing read-only subrepos from the monorepo. We'd like to remove/archive these read-only repos altogether, and the scripts that copy code into them.

As part of this, we need to update godog to reference the common modules (gherkin etc) via a path within the common repo, rather than the subrepo. See @aurelien-reeves's comment [here](https://github.com/cucumber/common/issues/1550#issuecomment-844766751).

I'm not all that familiar with the go ecosystem, but I can probably figure out a PR for this.

What I'm not sure is about the deployment process. If we make a release of godog with the new module paths, will end users have to upgrade if we then archive the subrepos?

aslakhellesoy:
Step one would be to see if we can get gherkin/go building by referencing messages from the monorepo.

mattwynne:
OK I've had a go at that [here](https://github.com/cucumber/common/pull/1608) but I could do with some help from someone who has some idea what they're doing with go. I have none.

mattwynne:
@vearutop any chance you could help me with this?

It's blocking a larger piece of work to decommission a bunch of obsolete scripts to maintain the mirror repos from cucumber/common.

vearutop:
Sure, let's do this after releasing v0.12.0.

vearutop:
I've created a couple of draft PR for this issue, one is a straightforward imports replace as suggested by this issue.
And another is a more complicated solution to embed dependencies into an internal packages of `godog`.

I don't have a strong opinion which approach is the best, both have pros and cons.

Let's decide which way to go and I can finish one of the drafts with documentation/changelog update.

mpkorstanje:
I think you have spotted the same problem I've been trying to deal with in Cucumber-JVM.

The messages and gherkin dependencies are exposed as part of cucumbers api.  This is made worse by the dependency structure in which messages is a dependency of everything. As a result updates to for example the html formatter often force a new major version.

In Cucumber-JVM I have chosen not to expose these dependencies. This was made easier by already having an existing API. This API uses an adapter pattern to shield users from rapid changes.

Is it possible do something similar in godog without also copying the dependencies to an internal module?

vearutop:
[...]

PR #404, opened 2021-07-07
Fix tests and CircleCI jobs

vearutop:
# Description

Current `main` fails CI jobs since recent merges.

# Motivation & context

This PR 
* addresses https://github.com/cucumber/godog/pull/402#issuecomment-874291502,
* applies `gofmt` formatting where missing,
* reverts minimal Go version back to 1.13 to align with CI images,
* fixes `go1.16` home directory configuration in CI,
* fixes obscure merge conflict where `MultiFormatter` was depending on obsolete (`v10` instead of `v16`) version of `messages-go`,
* fixes builder test with explicit `go mod tidy` to comply with [No automatic changes to go.mod and go.sum](https://blog.golang.org/go116-module-changes#TOC_3),
* removes parallelism in builder tests to avoid racy module updates.

## Type of change

## Note to other contributors

No notes.

Not required.

# Checklist:


vearutop:
Oh, actually the problem is bigger. Latest commit in `main` conflicts with the previous. Will fix.

internal/models/feature.go
```diff
@@ -38,7 +38,7 @@ func (f Feature) FindBackground(astScenarioID string) *messages.Background {
 		}
 	}
 
-	return bg
+	return nil
```
vearutop, line 41: This change reverts https://github.com/cucumber/godog/commit/5414f3c5da2714be31ee25ff0ad8c53ee3d21816#diff-11cc1058f443cce780d83b56496302454770d91e64cd1f228d2d45d38598e9a4R41.

@mbow could you please check if that change was intended?

run_test.go
```diff
@@ -350,7 +350,13 @@ func Test_FormatOutputRun(t *testing.T) {
...
 	)
 
-	file := filepath.Join(t.TempDir(), "result.xml")
+	dir := filepath.Join(os.TempDir(), t.Name())
+	err := os.MkdirAll(dir, 0755)
+	require.NoError(t, err)
+
+	defer os.RemoveAll(dir)
+
+	file := filepath.Join(dir, "result.xml")
```
vearutop, line 359: This change reverts usage of `t.TempDir()` to support go1.13 where the function is not available.

_examples/custom-formatter/go.mod
```diff
@@ -3,6 +3,10 @@ module custom-formatter
 go 1.14
 
 require (
-	github.com/cucumber/godog v0.11.0
+	github.com/cucumber/gherkin-go/v11 v11.0.0 // indirect
+	github.com/cucumber/godog v0.10.1-0.20210705192606-df8c6e49b40b
+	github.com/cucumber/messages-go/v10 v10.0.3 // indirect
 	github.com/cucumber/messages-go/v16 v16.0.1
+	github.com/hashicorp/go-immutable-radix v1.3.1 // indirect
+	github.com/hashicorp/go-memdb v1.3.2 // indirect
```
vearutop, line 11: This change pinpoints a not yet released version of `github.com/cucumber/godog` to resolve import dependency on `github.com/cucumber/messages-go/v16`.

mattwynne:
@vearutop I can't do any kind of meaningful review this as I don't know the codebase.

You have the commit bit, so if you're happy with I suggest you go ahead and merge.

If you have specific things you want feedback on first, please ask!

_examples/custom-formatter/go.mod
mattwynne: I don't want this PR to snowball, but is https://github.com/cucumber/godog/issues/389 worth bearing in mind here if we're changing the dependencies?

vearutop: As far as I understood, the idea of #389 is to start using `github.com/cucumber/common/messages/go/v16` packages instead of `github.com/cucumber/messages-go/v16`. 

I think such change should to be delivered in separate PR to change all relevant imports of the `github.com/cucumber/godog`.

Current change is only a fix in one of examples provided by `godog` which is organized as a separate module and so does not inherit parent (`godog`) module dependencies. Previously this example was using `godog` `v0.11.0`, but at the same was using `messages-go/v16` that was only supported by newer (untagged) version of `godog`. 

In this situation we have 3 options in my opinion:
* revert example to use `messages-go/v10` so that  it is again compatible with `godog` `v0.11.0`,
* update example to use pinned commit for newer `godog` (as done in this PR),
* remove `go.mod` from example, so that it inherits parent dependencies.

mattwynne (approved):

_examples/custom-formatter/go.mod
mattwynne: > As far as I understood, the idea of #389 is to start using github.com/cucumber/common/messages/go/v16 packages instead of github.com/cucumber/messages-go/v16.

That's right

I think such change should to be delivered in separate PR to change all relevant imports of the github.com/cucumber/godog.

👍

PR #428, opened 2021-08-27
Embed common packages from monorepo

### LINKED ISSUE #389: Update cucumber/common module references to use paths instead of subrepos

vearutop:
# Description

This PR adds a `make common` target to copy `messages` and `gherkin` packages from common monorepo into `godog`'s internal. This is an alternative solution to migrate away from `github.com/cucumber/messages-go` and `github.com/cucumber/gherkin-go` while not depending on a monorepo.

# Motivation & context

Once common monorepo has an update in `gherkin` and/or `messages`, a `godog` maintainer can run `make common` and create a PR with produced changes. 

Pros:
* more control for future breaking changes as `messages` package is not directly importable by end-users
* no runtime dependency on relatively large common monorepo

Cons:
* +10k lines of code in `godog` repo
* no way for users to import `internal/messages` in case they need it for advanced step definitions (can be solved by providing more local aliases in `godog` package)
* manual upgrade procedure for `common` packages (though it would still be manual when importing common directly)
* embedded code does not pass `golint` and is excluded from linting (may be fixed in monorepo, but with breraking changes in symbol names)
* overall code coverage drops when merged with embedded packages (can be fixed by improving coverage in monorepo)

Resolves #389.

## Type of change

- Breaking change (will cause existing functionality to not
  work as expected)

## Note to other contributors

No notes.

No update.

# Checklist:

- [ ] I have read the [**CONTRIBUTING**](../CONTRIBUTING.md) document.


vearutop:
Closing this, as we need to migrate to different standalone modules (https://github.com/cucumber/common/issues/2029).

PR #429, opened 2021-08-27
Use common packages from monorepo

### LINKED ISSUE #389: Update cucumber/common module references to use paths instead of subrepos

vearutop:
# Description

This PR changes imports of `messages` and `gherkin` to use common monorepo.

# Motivation & context

Pros:
* A small change to `godog`
* Code coverage and lint are not affected

Cons:
* Runtime double dependency on a large repo (one copy for `gherkin/go/v20` and one for `messages/go/v17`)
* Every further change of imports (e.g. to `messages/go/v18`) is technically a  breaking change for `godog` as end-users may depend on messages directly while implementing advanced step definitions.

Runtime dependency on a monorepo seems like not a big deal, for example in GitHub Actions those packages are retrieved in 1-2 seconds.

Resolves #389.

## Type of change

- Breaking change (will cause existing functionality to not
  work as expected)

## Note to other contributors

No notes.

No update.

# Checklist:

- [ ] I have read the [**CONTRIBUTING**](../CONTRIBUTING.md) document.

### DIFF
--- a/_examples/custom-formatter/go.sum
+++ b/_examples/custom-formatter/go.sum
@@ -32,11 +32,11 @@ github.com/coreos/go-systemd v0.0.0-20190321100706-95778dfbb74e/go.mod h1:F5haX7
 github.com/coreos/pkg v0.0.0-20180928190104-399ea9e2e55f/go.mod h1:E3G3o1h8I7cfcXa63jLwjI0eiQQMgzzUDFVpN/nH/eA=
 github.com/cpuguy83/go-md2man/v2 v2.0.0/go.mod h1:maD7wRr/U5Z6m/iR4s+kqSMx2CaBsrgA7czyZG/E6dU=
 github.com/creack/pty v1.1.9/go.mod h1:oKZEueFk5CKHvIhNR5MUki03XCEU+Q6VDXinZuGJ33E=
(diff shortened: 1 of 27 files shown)


mattwynne:
@vearutop Sorry it's taken me so long to give you feedback on this and #428.

One bit of context that might affect this is that we've decided to [break up the monorepo](https://github.com/cucumber/common/issues/1724), so the burden on users of downloading the dependencies will be much smaller once that's done.

Do you have a favourite of the two?

mattwynne:
Note that [messages](https://github.com/cucumber/common/issues/2029) is well on the way to being removed.

vearutop:
Closing this, as we need to migrate to different standalone modules (https://github.com/cucumber/common/issues/2029).

PR #437, opened 2021-10-20
Automate binary releases

vearutop:
# Description

This PR adds GitHub Action to build and upload binary artifacts on release creation.
Also it arranges examples in a single module that references current `godog`, so that examples are always up to date with latest `godog`.

# Motivation & context

Currently artifacts uploading is manual and makes releases more effortful than they should be. 
With this automation we can reduce the burden for smaller releases and tag every meaningful change as soon as it is ready and PR is merged.

## Type of change

## Note to other contributors

`godog.Version` is now populated from module info and does not need manual editing. 

No update.

# Checklist:


mattwynne:
> This looks great!
> 
> @mattwynne @aurelien-reeves let’s discuss this to make sure it’s aligned with the general release process that is described in https://github.com/cucumber/.github/blob/main/RELEASING.md

Sure, it looks good to me.

@vearutop did you consider using https://github.com/goreleaser/goreleaser?

vearutop:
@aslakhellesoy I think `godog` deviates from RELEASING.md guidelines in a way that `release/*` branches are not used since `v0.11.0` and also it seems there were no automations configured for these branches. We also don't backport features or fixes to older versions.

Collecting unreleased changes and then eventually (or periodically) releasing them seems unnecessary limiting for small self-contained changes, though it totally makes sense if multiple changes need to be shipped and tested together.

I think we can make patch version releases asap with only CHANGELOG.md updates, and prepare more documentation in `release-notes` for major releases (having breaking changes).  Reducing release cycle (especially for bug fixes) may help with adoption and faster feedback.

@mattwynne I was conservatively thinking to make this with least changes and dependencies. We can upgrade to `goreleaser`, it has interesting additional features like brew formula management though I haven't used `goreleaser` in the past.

The approach from this PR was tested with some of my personal [projects](https://github.com/vearutop/plt/releases).

aslakhellesoy:
> @aslakhellesoy I think `godog` deviates from RELEASING.md guidelines in a way that `release/*` branches are not used since `v0.11.0` and also it seems there were no automations configured for these branches. We also don't backport features or fixes to older versions.
> 

The reason we use `release/*` branches is purely to limit who can make a release. To do this we configure each repo so that only members of the @cucumber/release team can push to these branches. We have GitHub actions that are only triggered on a push to `release/*`, and also GitHub secrets that are only available to GitHub Actions on `release/*` branches.

The `release/*` branches in this release process should be thought of as ephemeral - they should ideally be removed after the release (we just haven't automated that part yet).

You can see how this is set up in the tag-expressions repo:
* https://github.com/cucumber/cucumber-expressions/settings/branches
* https://github.com/cucumber/cucumber-expressions/settings/environments/302175658/edit

mattwynne:
I suggest we merge this, and we can deal with the other two suggestions (goreleaser, automation via `release/*` branches) as later iterations.

mattwynne:
What's the new `mod_version.go` file for? Even though it's simple it's not covered by tests - should we add a unit test for it?

vearutop:
`mod_version.go` is version info initializer that leverages Go Module information that is injected during the build. It is a separate file to only compile with go1.12 or later. It works somewhat similar to LDFLAGS, but automatically. When user runs `go install github.com/cucumber/godog/cmd/godog@latest`, `Version` value will be populated with actual version used in the build (e.g. `v0.12.3`).

We can remove build tags and move the contents to an existing file given that go1.11 is very old and is not supported anyway.

I'll add a unit test for this file to improve coverage.

Issue #406, opened 2021-07-09
WANTED: More regular maintainers 🔧🙋‍♀️🙋🏾‍♂️🙋🏼 


mattwynne:
Since Cucumber adopted this project, the bulk of the maintenance effort has fallen to just [one or two individuals](https://github.com/cucumber/godog/graphs/contributors).

While we're super-grateful to those individuals, this situation puts too much pressure on those individuals and is unsustainable. Right now, with those people busy with other stuff, pull requests and issues are starting to pile up. Thanks to @vearutop who's already started to [step in and help](https://github.com/cucumber/godog/commit/9a335aed221fadf59b68323742f2189f4a5d782b). 🙏🏻

It's clear from the number of GitHub stars that lots of people are using this project. I'm raising this issue as a call to the Go community to see who would like to get more involved.

Some reasons to step forward:

* Be a part of a welcoming community of developers who take test-driven development for granted. 🚀🎉
* Enjoy the chance to pair with and be mentored by people like @aslakhellesoy, @aurelien-reeves and I who have years of experience to share. 🤔🤝
* Know that the Cucumber community has an explicit commitment to diversity, [equity](https://www.mentalfloss.com/article/625404/equity-vs-equality-what-is-the-difference) and inclusion and that if you're from an under-represented minority in tech, you'll be welcome here. ✊🏿🏳️‍🌈🏳️‍⚧️❤️

Here are some ways you can help the project:

* [Triage](https://github.com/cucumber/common/blob/main/TRIAGE.md) existing issues
* Review/triage other people's pull requests
* Submit pull requests that (at least partially) solve an issue.
* Merge pull requests (once you have the commit bit)
* Join community discussions about [bigger strategic changes](https://github.com/go-bdd/gobdd/issues/137)

All of these are helpful and very much appreciated.

We have limited Go fluency in the team of regular maintainers at the moment, but we're here to help and available to pair with you if you need help figuring something out.

Please let us know your thoughts in the comments, or come for a chat in the [Community Slack](https://cucumber.io/community#slack) in the [#commiters-go](https://cucumberbdd.slack.com/archives/CA5NJPDJ4) channel.

gnoe:
I'll love to contribute to this project!

cpustejovsky:
I'm happy to contribute to this project! I fell in love with BDD and Cucumber on my last project and would like to do what I can to help godog

soheilrt:
[...]

PR #439, opened 2021-11-12
Friendlier contributing guide

mattwynne:
A baby step towards #406.

Having paired with a new contributor on godog this week, I noticed how we don't really document how to hack on godog.

I also noticed how unwelcoming the CONTRIBUTING.md file was.

I've changed the tone to be warmer and more welcoming, and I've also made an attempt to explain what people need to do if they want to start hacking on godog.

It would be amazing to have some input here from others who have more familiarity with this codebase. For example, I noticed there's a `Makefile` with a `make test` target, but to run that I need to install golint which is deprecated. Should we still advise people to install that too, and use the `make test` target?

*Update* I figured out how to install `golint` so I've changed the instructions to model that.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -23,6 +23,6 @@ Once that's done, try running the tests:
 
 If everything passes, you're ready to hack!
 
-[install go](https://golang.org/doc/install)
+[install go]: https://golang.org/doc/install
 [community Slack]: https://cucumber.io/community#slack
 [raise an issue]: https://github.com/cucumber/godog/issues/new/choose
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -11,11 +11,7 @@ You can chat with us in the [#committers-go](https://cucumberbdd.slack.com/archi
 
 ## Setup
 
-To get your development environment set up, you'll need to [install Go].
-
-Now install golint:
-
-    go install golang.org/x/lint/golint@latest
+To get your development environment set up, you'll need to [install Go]. We're currently using version 1.17 for development.
 
 Once that's done, try running the tests:
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -19,7 +19,7 @@ Now install golint:
 
 Once that's done, try running the tests:
 
-    make
+    make test
 
 If everything passes, you're ready to hack!
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,17 +1,24 @@
-## About to contribute?
+# Welcome 💖
 
-We appreciate that. But before you do, please learn our basic rules:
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -1,17 +1,28 @@
...
+
+    go install golang.org/x/lint/golint@latest
+
+Once that's done, try running the tests:
+
+    make
+
+If everything passes, you're ready to hack!
+
+[install go](https://golang.org/doc/install)
```
aurelien-reeves, line 26:
```suggestion
[install go]: https://golang.org/doc/install
```

README.md
```diff
@@ -602,3 +602,5 @@ A simple example can be [found here](/_examples/custom-formatter).
 [behat]: http://docs.behat.org/ "Behavior driven development framework for PHP"
 [cucumber]: https://cucumber.io/ "Behavior driven development framework"
 [license]: https://en.wikipedia.org/wiki/MIT_License "The MIT license"
+[contributing guide]: [https://github.com/cucumber/godog/blob/main/CONTRIBUTING.md]
```
aurelien-reeves, line 605:
```suggestion
[contributing guide]: https://github.com/cucumber/godog/blob/main/CONTRIBUTING.md
```

CONTRIBUTING.md
```diff
@@ -1,17 +1,28 @@
...
+
+To get your development environment set up, you'll need to [install Go].
+
+Now install golint:
+
+    go install golang.org/x/lint/golint@latest
+
+Once that's done, try running the tests:
+
+    make
```
vearutop, line 22: Nit: more explicit command to avoid accidental change of behavior in future.

```suggestion
    make test
```

CONTRIBUTING.md
```diff
@@ -1,17 +1,28 @@
...
+
+You can chat with us in the [#committers-go](https://cucumberbdd.slack.com/archives/CA5NJPDJ4) channel in our [community Slack], or feel free to [raise an issue] if you're experiencing any friction trying make your contribution.
+
+## Setup
+
+To get your development environment set up, you'll need to [install Go].
+
+Now install golint:
+
+    go install golang.org/x/lint/golint@latest
```
vearutop, line 18: Btw, we can update `Makefile` to use 
```
go run golang.org/x/lint/golint@latest github.com/cucumber/godog
```
instead of
```
golint github.com/cucumber/godog
```
and avoid mentioning `golint` in setup.

Performance difference is negligible, caveat is that such `go run` is only available since `go1.17` (though `go1.16` will become unsupported in beginning of 2022 with release of `go1.18`).

vearutop (approved):

CONTRIBUTING.md
mattwynne: Good plan! Fixed in 8d3bee2deba17ff4613afa6dd393e919ff951460

Issue #440, opened 2021-11-16
Rule is not supported


michaelsauter:
### What version of godog are you using?
```
> godog version: 0.12.2
```

### What version of Go are you using?
```
> go version: 1.16
```

### What did you do?
Given a feature file like the following:

```
Feature: simple feature
  simple feature description

  Rule: simple rule
    Example: simple example
```

Running `godog` will panic:

```
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
        panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x70 pc=0x11cc0c6]

goroutine 34 [running]:
github.com/cucumber/godog/internal/formatters.(*Pretty).printStep(0xc0002020d0, 0xc000250360, 0xc000208680)
        /Users/michaelsauter/go/pkg/mod/github.com/cucumber/godog@v0.12.2/internal/formatters/fmt_pretty.go:384 +0xf26
``` 

### Additional context
I've looked a bit into the source code and added the following test to `pretty.feature`:
```
  Scenario: Support of Feature Plus Rule Node
    Given a feature "features/simple.feature" file:
    """
        Feature: simple feature
            simple feature description

        Rule: simple rule
            Scenario: simple scenario
                simple scenario description
    """
    When I run feature suite with formatter "pretty"
    Then the rendered output will be as follows:
    """
      Feature: simple feature
        simple feature description

        Rule: simple rule # features/simple.feature:3

        Scenario: simple scenario # features/simple.feature:4

      1 scenarios (1 undefined)
      No steps
      0s
    """
```

This lead me to apply the following fix in `fmt_pretty.go`:
```go
func (f *Pretty) scenarioLengths(pickle *messages.Pickle) (scenarioHeaderLength int, maxLength int) {
	feature := f.Storage.MustGetFeature(pickle.Uri)
	astRule := feature.FindRule(pickle.AstNodeIds[0])
	astScenario := feature.FindScenario(pickle.AstNodeIds[0])
	astBackground := feature.FindBackground(pickle.AstNodeIds[0])

	if astScenario != nil {
		scenarioHeaderLength = f.lengthPickle(astScenario.Keyword, astScenario.Name)
		maxLength = f.longestStep(astScenario.Steps, scenarioHeaderLength)
	}

	if astBackground != nil {
		maxLength = f.longestStep(astBackground.Steps, maxLength)
	}
	if astRule != nil {
[...]

PR #441, opened 2021-11-16
Support rule keyword

michaelsauter:
Related to #440.

After reading a bit more in the code I realized my original understanding of how it worked was wrong. Supporting `Rule` might not be that hard after all. I started implementing it a little. This is nowhere near finished though ...


mxygem:
@michaelsauter - Thanks for running with this!

mattwynne:
Awesome @michaelsauter and thanks for doing your work in public!

Let me know if you want to do some pairing ⚡ you can book a slot with me [here](https://calendly.com/mattwynne).

michaelsauter:
Thanks!

Sorry I did not get to it again, but I should find some time in a couple of days. Will update here and might come back to your offer of pairing!

michaelsauter:
@mxygem @mattwynne Alright, I finally had the time to work a bit on this. It's now at a state where at least the pretty format seems to work well. I have added some tests that demonstrate the behaviour. I would love to hear some feedback, and if this is still missing anything.

michaelsauter:
I'll have another look into the code coverage when I get to it.

features/formatter/pretty.feature
```diff
@@ -23,6 +23,181 @@ Feature: pretty formatter
...
+            simple scenario description
+    """
+    When I run feature suite with formatter "pretty"
+    Then the rendered output will be as follows:
+    """
+      Feature: simple feature
+        simple feature description
+
+        Background: simple background
+            Given passing step
```
mattwynne, line 44: I'm confused by this indentation - do we indent the steps by 4 characters within a Background?

mattwynne: Or is this about having a universal fixed indentation for every step irrespective of whether it's inside a Rule, Background, Scenario etc?

mattwynne:
Great work @michaelsauter. It looks like maybe we need some unit tests to cover the details of the `Find*` methods on `feature.go`. I'm surprised those aren't being picked up from the feature tests though - @vearutop are the godog feature tests included in the code coverage?

mattwynne:
One thing I wonder about might be missing, but we could do this in a separate PR, is to add support for tagging rules.

vearutop:
Yeah, feature tests are included in code coverage, apparently current tests are not triggering particular code branches.

features/formatter/pretty.feature
michaelsauter: I think the indentation is actually ignored in the tests. I changed indentation and it did not caused the tests to fail.

michaelsauter:
I think there are a few more combinations that I can add to see if I can trigger those other branches. That said, I already added tests to cover some branches that were not tested before so I am a bit confused about the code coverage.

> One thing I wonder about might be missing, but we could do this in a separate PR, is to add support for tagging rules.

Good point, will check how hard this is and see if I can add it to this PR.

Sorry it might be another couple of days before I get back to this ...

azuk-bread:
@michaelsauter any updates on this one?  Could really use it :)

vearutop:
If @michaelsauter does not have capacity to finish with tests and code style, I think we can merge this PR to a temporary branch and I can then add missing parts. Let's wait couple of days for response.

michaelsauter:
Sorry, I lost track of this and currently do not have a need for this. Would appreciate if you take this over. Thanks!

mattwynne:
Superseded by #480. Thanks @dumpsterfireproject!

PR #445, opened 2021-12-03
Use more expressive names

mattwynne:
# Description

Rename some terse variable names to be more expressive.

# Motivation & context

I'm reading the suite code, which is the heart of godog's execution model, and I find it hard to read the code when I look at variables named `s` or `sr`. I prefer something a little bit longer and more expressive.

I guess this may not be idiomatic Go? I'm ready to hear feedback on that!

I don't think this has broken any public APIs, but please let me know if it has.

## Type of change

# Checklist:


internal/models/stepdef.go
```diff
@@ -36,10 +36,10 @@ type StepDefinition struct {
 var typeOfContext = reflect.TypeOf((*context.Context)(nil)).Elem()
 
 // Run a step with the matched arguments using reflect
-func (sd *StepDefinition) Run(ctx context.Context) (context.Context, interface{}) {
+func (stepDef *StepDefinition) Run(ctx context.Context) (context.Context, interface{}) {
```
mpkorstanje, line 39: What's a `def`?

mxygem:
> I guess this may not be idiomatic Go? I'm ready to hear feedback on that!

You're correct here. Since method receivers are generally very important things in their methods, a one letter assignment is common. The other place where single letters are generally suggested are within things like loops/ranges. [This post](https://dave.cheney.net/practical-go/presentations/qcon-china.html#_use_a_consistent_naming_style) is a **fantastic** read. I linked to a particular section on naming and what's most pertinent to this topic is the first `tip` comment within it.

internal/models/stepdef.go
mxygem: Given my other comment below on method receivers, I'd personally prefer this be kept as at least `sd` if not fully changed to `s`.

mxygem: Also to piggyback on @mpkorstanje 's comment, if I'm assuming their intent correctly, the file could probably use being renamed to `stepdefinition.go`

suite.go
```diff
@@ -43,8 +43,8 @@ const (
 	StepPending = models.Pending
 )
 
-type suite struct {
```
mxygem, line 46: Not necessarily against this change though I'm curious what prompted this particular change? IMO, naming something with a prefix of `test` _can_ denote a version of something used in tests. While it's true technically the suite is for testing is that something that one would imply? I suppose given the fact that you changed it, the answer to my last question is no? 😅

suite.go
```diff
@@ -109,87 +109,87 @@ func (s *suite) runStep(ctx context.Context, pickle *Scenario, step *Step, prevS
...
 	return ctx, err
 }
 
-func (s *suite) runBeforeStepHooks(ctx context.Context, step *Step, err error) (context.Context, error) {
+func (suite *testSuite) runBeforeStepHooks(ctx context.Context, step *Step, err error) (context.Context, error) {
 	hooksFailed := false
 
-	for _, f := range s.beforeStepHandlers {
-		hctx, herr := f(ctx, step)
+	for _, handler := range suite.beforeStepHandlers {
```
mxygem, line 191: This particular usage like others isn't _wrong_ as it's still valid go, though it does qualify as against styling. I'd also add that `handler` is only used once within the loop on the immediately following line, could be assumed that it is a handler given we're ranging over something called `beforeStepHandlers`, and it also breaks the consistency of the following `hctx` & `herr` for handler context and handler error respectively.

suite.go
```diff
@@ -212,9 +212,9 @@ func (s *suite) runBeforeStepHooks(ctx context.Context, step *Step, err error) (
 	return ctx, err
 }
 
-func (s *suite) runAfterStepHooks(ctx context.Context, step *Step, status StepResultStatus, err error) (context.Context, error) {
-	for _, f := range s.afterStepHandlers {
-		hctx, herr := f(ctx, step, status, err)
+func (suite *testSuite) runAfterStepHooks(ctx context.Context, step *Step, status StepResultStatus, err error) (context.Context, error) {
+	for _, handler := range suite.afterStepHandlers {
+		hctx, herr := handler(ctx, step, status, err)
```
mxygem, line 217: Hmm. Scrolling down to this example also seems to muddy things a little further. What also adds more to the confusion is that what are called handlers here are actually hooks, so there's an odd naming swap. I haven't looked deep enough to see if there's a particular reason why though.

When trying to understand just what calling `handler` did, it took a little more digging than would probably be good for others working on the code. I wonder if there's a slightly different way of organizing things where we could change Hook to be a type/struct or even an interface so that we could do something like, `Hook.run()`? That seems to read a little better anyhow.

suite.go
vearutop: There is indeed some historical inconsistency in names related to hooks, I think it would be beneficial to use word `hook` instead of `handler` where hook semantics is implied (e.g. rename `beforeStepHandlers`). 

Range variable `f` denotes function, I think `h` might be a better name.

suite.go
mattwynne: It was simply to get the `suite` name out of the way so I could use it for the receiver. But if we're going back to `s` for the receiver then we don't need it.

suite.go
mattwynne: One reason I wanted to use a longer name here was because I missed the most important line (where the hook function is called) initially when skimming the code.

suite.go
mattwynne: > > I guess this may not be idiomatic Go? I'm ready to hear feedback on that!
> 
> You're correct here. Since method receivers are generally very important things in their methods, a one letter assignment is common. The other place where single letters are generally suggested are within things like loops/ranges. [This post](https://dave.cheney.net/practical-go/presentations/qcon-china.html#_use_a_consistent_naming_style) is a **fantastic** read. I linked to a particular section on naming and what's most pertinent to this topic is the first `tip` comment within it.

Thanks! These terse names make more sense now. I definitely want the code to be idiomatic, so I'll change these back.

mattwynne:
> > I guess this may not be idiomatic Go? I'm ready to hear feedback on that!
>
> You're correct here. Since method receivers are generally very important things in their methods, a one letter assignment is common. The other place where single letters are generally suggested are within things like loops/ranges. This post is a fantastic read. I linked to a particular section on naming and what's most pertinent to this topic is the first tip comment within it.

Thanks! These terse names make more sense now. I definitely want the code to be idiomatic, so I'll change these back.

suite.go
mattwynne: Resolved in 9297c4d6b28bada946c3e8ebf85be3ee87fb52c2

suite.go
mattwynne: Ideally I would like to see an abstraction wrapped around both the step definitions and the hooks that lets us treat all of them as "test steps" which we can just compile into one big list and execute in turn. But I think that's for another PR!

internal/models/stepdef.go
mattwynne: Good points both - done!

suite.go
mattwynne: I've renamed it to `hook` but we can use `h` if you think that's more idiomatic.

mattwynne: @vearutop I've also renamed `*handler` to `*hook` everywhere. Better?

suite.go
```diff
@@ -109,87 +106,85 @@ func (s *suite) runStep(ctx context.Context, pickle *Scenario, step *Step, prevS
...
+		stepResult.Status = models.Failed
+		s.storage.MustInsertPickleStepResult(stepResult)
 
 		return ctx, err
 	}
 
 	if ctx, undef, err := s.maybeUndefined(ctx, step.Text, step.Argument); err != nil {
 		return ctx, err
 	} else if len(undef) > 0 {
-		if match != nil {
```
vearutop, line 153: I think this rename takes away some of the context.

The fact that this var is a step definition can be derived from its type, so the name does not convey additional information. However this var holds a specific step definition: a particular match.

I think it would be better to keep original name to reflect the purpose rather than content.

internal/models/stepDefinition.go
```diff
@@ -50,15 +50,15 @@ func (sd *StepDefinition) Run(ctx context.Context) (context.Context, interface{}
...
+	if len(s.Args) < numIn {
+		return ctx, fmt.Errorf("%w: expected %d arguments, matched %d from step", ErrUnmatchedStepArgumentNumber, typ.NumIn(), len(s.Args))
 	}
 
 	for i := 0; i < numIn; i++ {
 		param := typ.In(i + ctxOffset)
 		switch param.Kind() {
 		case reflect.Int:
-			s, err := sd.shouldBeString(i)
+			s, err := s.shouldBeString(i)
```
vearutop, line 61: This rename introduces slight visual ambiguity by shadowing variable with different type and purpose. In this particular case it is not big deal as new scope (that uses `s` string) if very small and is hard to misinterpret, in larger scopes that could be considered a questionable practice for higher chance of confusion.

suite.go
```diff
@@ -57,25 +57,24 @@ type suite struct {
...
 
-	// suite event handlers
-	beforeScenarioHandlers []BeforeScenarioHook
-	beforeStepHandlers     []BeforeStepHook
-	afterStepHandlers      []AfterStepHook
-	afterScenarioHandlers  []AfterScenarioHook
+	beforeScenarioHooks []BeforeScenarioHook
+	beforeStepHooks     []BeforeStepHook
+	afterStepHooks      []AfterStepHook
+	afterScenarioHooks  []AfterScenarioHook
```
vearutop, line 63: 👍

suite.go
mattwynne: True. And would `matchingStepDefinition` be too long-winded a name?

vearutop: It is fine for me, though in general shorter names are preferred as long as they tell enough and don't bring confusion (e.g. having two different matches of different kinds in the same scope would require more specific names). 

https://research.swtch.com/names

mattwynne: Out of interest, what does that `undef` variable there above contain? Is that the list of undefined, _nested_, steps?

vearutop: I checked using tests, it holds a list of one item, a string with step text, e.g. `another undefined step`.

mattwynne:
I'm gonna abandon this branch - it's got out of date. I appreciate the feedback @mxygem and @vearutop - I learned a good deal from it. I may come back with smaller more focssed PRs salvaging some of the more useful commits from this.

mxygem:
@mattwynne No worries! Is there something that I missed or could have done to help? While I know you don’t necessarily need to learn all about Go, but this lovely open source book is something I recommend to basically everyone regardless of skill: [Learn Go With Tests](https://quii.gitbook.io/learn-go-with-tests/)

mattwynne:
> @mattwynne No worries! Is there something that I missed or could have done to help?

Not at all! You gave great feedback. I just dropped the ball and, with some distance, the changes I'd made seemed not worth fighting through the merge conflicts.

> While I know you don’t necessarily need to learn all about Go, but this lovely open source book is something I recommend to basically everyone regardless of skill: [Learn Go With Tests](https://quii.gitbook.io/learn-go-with-tests/)

Thanks for the tip - bookmarked!

PR #457, opened 2022-02-10
Switch from golint to staticcheck

### LINKED ISSUE #448: Switch from go lint to staticcheck.io
Since go lint has been deprecated (https://github.com/golang/go/issues/38968), it would be good to move to an actively maintained static checker for Go. [staticcheck.io](https://staticcheck.io/) comes [highly recommended](https://groups.google.com/g/golang-nuts/c/rCP70Aq_tBc?pli=1) and is sponsored by Google. 

I would like to submit a PR in the future to implement this, but would like some feedback on whether this would be appreciated before I start.

carleeto:
# Description

This PR switches static checking from the deprecated tool golint to staticcheck

# Motivation & context

It switches static linting from a deprecated tool to a recommended one.

Fixes #448 

## Type of change

# Checklist:

- [ ] My change needed additional tests
- [x] My change requires a change to the documentation. (doesn't require a change, but I have changed CONTRIBUTING.md)
  - [x] I have updated the documentation accordingly.  (to make it easier for future authors who update dependencies)


carleeto:
I've enabled staticcheck in CI. There are a few errors. 
Before doing this, go vet was failing on go1.16. 

Plan:
First, get staticcheck back to green by disabling failing checks.
Next, understand why go vet failed on go 1.16
Finally, ensure CI is green and reopen the PR

mattwynne:
I'm currently pairing with @carleeto on this. It's tricky for him to get feedback as I have to press the "Approve" button each time he pushes! I'm confident he's going to be a useful member of the team so I've pre-empted the commit-bit bot and added him to the `cuucmber/committers` team manually.

vearutop:
Maybe let's try downloading `staticcheck` from https://github.com/dominikh/go-tools/releases/download/2021.1.2/staticcheck_linux_amd64.tar.gz instead of building it, it might be faster and less invasive.

mattwynne:
We seem to have a weird issue where the `go vet` command is failing in CI complaining about a missing `go.sum` entry which we don't think is missing - any idea why that would be happening @vearutop?

We've tried reproducing locally on a v1.16 go and it seems to work fine.

.github/workflows/test.yml
```diff
@@ -28,11 +28,13 @@ jobs:
...
       - name: Run gofmt
         run: gofmt -d -e . 2>&1 | tee outfile && test -z "$(cat outfile)" && rm outfile
-      - name: Run golint
+      - name: Run staticcheck
         run: |
-          go install golang.org/x/lint/golint@latest
-          golint -set_exit_status ./...
-          cd _examples && golint -set_exit_status ./... && cd ..
+          go run honnef.co/go/tools/cmd/staticcheck github.com/cucumber/godog
+          go run honnef.co/go/tools/cmd/staticcheck github.com/cucumber/godog/cmd/godog
```
vearutop, line 34:
```suggestion
          go run honnef.co/go/tools/cmd/staticcheck@latest github.com/cucumber/godog
          go run honnef.co/go/tools/cmd/staticcheck@latest github.com/cucumber/godog/cmd/godog
```

Let's try `@latest`, I think it should run without modifying local `go.mod` / `go.sum`.

Makefile
```diff
@@ -12,8 +12,8 @@ test: check-go-version
 	@echo "running all tests"
 	@go install ./...
 	@go fmt ./...
-	@go run golang.org/x/lint/golint@latest github.com/cucumber/godog
-	@go run golang.org/x/lint/golint@latest github.com/cucumber/godog/cmd/godog
+	@go run honnef.co/go/tools/cmd/staticcheck github.com/cucumber/godog
+	@go run honnef.co/go/tools/cmd/staticcheck github.com/cucumber/godog/cmd/godog
```
vearutop, line 16:
```suggestion
	@go run honnef.co/go/tools/cmd/staticcheck@latest github.com/cucumber/godog
	@go run honnef.co/go/tools/cmd/staticcheck@latest github.com/cucumber/godog/cmd/godog
```

.github/workflows/test.yml
```diff
@@ -28,11 +28,13 @@ jobs:
             ${{ runner.os }}-go-cache
       - name: Run gofmt
         run: gofmt -d -e . 2>&1 | tee outfile && test -z "$(cat outfile)" && rm outfile
-      - name: Run golint
+      - name: Run staticcheck
```
vearutop, line 31:
```suggestion
      - name: Run staticcheck
        if: matrix.go-version == '1.17.x'
```

.github/workflows/test.yml
mattwynne: We talked about that, but we were thinking it could be problematic if they released a new version of staticcheck with new rules and our build started to fail the next time someone pushed and our CI downloaded the new version.

Does that make sense?

vearutop: Yes, sure, let's use fixed version then
```
go run honnef.co/go/tools/cmd/staticcheck@v0.2.2 github.com/cucumber/godog
```

`go run` works slightly different with and without version suffix.

mattwynne: @vearutop are you thinking it's better if we keep `go.mod` / `go.sum` for _runtime_ dependencies, as opposed to for developer tools?

carleeto: Will try running with the fixed version suffix

vearutop: Yes, I think versioning CI tools with `go.mod` is a bit of an abuse, it is not a bad thing, but it grows dependency graph of a module and may potentially affect library consumers as a transitive dependency.

My favorite approach for CI tools is to download pre-built binaries of them.

carleeto: Where would you like the staticcheck binary stored?

carleeto:
@vearutop, @mattwynne: go vet checks in the 1.16 build are failing because the _examples folder is its own submodule. I'd like to understand - is this intentional?

carleeto:
> @vearutop, @mattwynne: go vet checks in the 1.16 build are failing because the _examples folder is its own Go module (the folder has its own go.mod and go sum files - https://github.com/cucumber/godog/blob/main/_examples/go.mod). I'd like to understand the reasoning behind this.

I think I understand why - its an example of how a user would use godog, so it needed to be its own sub-module. Am I right?

mdelapenya:
Have you thought using https://pre-commit.com and handle the static checks there? Then the same checks would be done at CI level than at local one

carleeto:
I started with tools.go and switched to storing a binary based on a suggestion by @vearutop. 

Personally, I prefer the tools.go approach. That said, if I could get a decision on the approach to be used, I'll go with it.

mdelapenya:
> If I understand correctly, the intent is to run a pre-commit check so that code that is committed is well formed.

The intend would be running the very same "hooks" across any member of the team including contributors. And a hook could be running whatever thing: we use it to check your shell scripts format, yml indent... Or even run the unit test if needed (depending of course the amount of time they take).

> 
> If that's the case, then the same can be accomplished using vanilla .githooks and Go.
> 
> From what I understand, pre-commit.com would add a python dependency, which doesn't really seem to add value.

mattwynne:
@vearutop do you have any comments on the way Carl's include the binary for static check - was that what you meant, or did you mean to script a download of the binary during the build process?

vearutop (approved):
Looks good to me 👍

vearutop:
> go vet checks in the 1.16 build are failing because the _examples folder is its own Go module (the folder has its own go.mod and go sum files - https://github.com/cucumber/godog/blob/main/_examples/go.mod). I'd like to understand the reasoning behind this.

Examples try to represent realistic cases and for that they have extra dependencies (for example `github.com/go-sql-driver/mysql`), these dependencies are redundant for `godog` itself, that's the reason why examples have their own module.

[In the past](https://github.com/cucumber/godog/blob/v0.12.0/_examples/assert-godogs/go.mod) each example had its own module and a dependency on already released version of `godog`, but that was not very convenient because examples had to be updated separately from `godog`.

mattwynne (approved):

mattwynne:
Thanks @carleeto and welcome to the Godog contributors team! 🎉

carleeto:
Cheers @mattwynne. Thanks @vearutop and @mdelapenya for the review too.

PR #476, opened 2022-05-13
add new option for created features with parsing from byte slices

akaswenwilk:
Hello!  In my work we are trying to build in a pre-processor for our feature files so we can have some dynamic feature generation when running our tests.  I made this for us to use but though maybe it could be interesting for more people than just us.  I've added another option to define custom features to be used in the suite, and then added an interface to pass in a map[string][]byte representing the path/read contents from the file (in our case, the contents after doing preprocessing).  I think this implementation is maybe not fully working since it seems to be missing the line filtering, but other than that I think I've got all the other inner workings there.  Just wanted to submit it if it could be interesting for anyone!

Leave me a comment if you think this is a good idea and then I can try adding some testing/additional documentation

- :bank: Refactoring/debt/DX (improvement to code design, tooling, documentation etc. without changing behaviour)
- :boom: Breaking change (incompatible changes to the API)

### 📋 Checklist:

----


vearutop:
This is an interesting feature and I think it can be very helpful to improve portability/embedability of tests. Let me check it in more detail to see if the API of this can be improved to be more accessible.

mattwynne:
Thanks for sharing! Can you add some tests for this @akaswenwilk?

akaswenwilk:
> Thanks for sharing! Can you add some tests for this @akaswenwilk?

Sure thing!

akaswenwilk:
hi @mattwynne , so sorry for the delay to get back to this.  I've added some tests for the ParseFromBytes function in the parser, but now I'm trying to address the interface a bit.  Currently I've added a Features option where someone can manually pass in a slice of features, but maybe actually it makes more sense to simple add a different option instead (like FeatureContents) which would be a map[string][]byte of all the features to be run.  Otherwise the user has to generate a slice of Feature models using the RetrieveFeaturesFromBytes method.  Any thoughts on which might be nicer?  At which point I can try adding a test in run_test.go and maybe we can finalize this? Thanks for your help!

akaswenwilk:
@vearutop, sorry for the delay, but how about this for improved API?  This way the user can just pass in a map[string][]bytes directly as an option and not have to deal with trying to first parse their bytes into a feature to pass in as an option.  Could be a bit simpler?

internal/flags/options.go
```diff
@@ -64,4 +64,9 @@ type Options struct {
 
 	// TestingT runs scenarios as subtests.
 	TestingT *testing.T
+
+	// FeatureContents allows passing in each feature manually
+	// where the contents of each feature is stored as a byte slice
+	// in a map entry
+	FeatureContents map[string][]byte
```
vearutop, line 71: Let's make this a slice of structs to improve clarity of data items.
Something like.
```
	// Features allow passing in each feature manually
	// where the contents of each feature is stored as a byte slice
	// in a map entry
	Features []Feature
}

// Feature contains path and contents of a gherkin file.
type Feature struct {
	Path    string
	Content []byte
}
```

run.go
```diff
@@ -227,7 +227,12 @@ func runWithOptions(suiteName string, runner runner, opt Options) int {
 	runner.fmt = multiFmt.FormatterFunc(suiteName, output)
 
 	var err error
-	if runner.features, err = parser.ParseFeatures(opt.Tags, opt.Paths); err != nil {
+	if len(opt.FeatureContents) > 0 {
+		runner.features, err = parser.ParseFromBytes(opt.Tags, opt.FeatureContents)
+	} else {
+		runner.features, err = parser.ParseFeatures(opt.Tags, opt.Paths)
+	}
```
vearutop, line 234: Let's change logic here so that both `Paths` and bytes can work together. We can check for non-empty `opt.Paths` and append `runner.features`.

vearutop:
@akaswenwilk overall looks good to me (apart from 2 minor suggestions that I left). 👍

wichert:
I'm wondering if it makes sense to provide an ability to pass in an `fs.FS` instance, along with a wrapper to expose your map as something that implements FS.

vearutop:
I think `fs.FS` does not have enough info (you need to know names in advance).
```go
// An FS provides access to a hierarchical file system.
//
// The FS interface is the minimum implementation required of the file system.
// A file system may implement additional interfaces,
// such as ReadFileFS, to provide additional or optimized functionality.
type FS interface {
	// Open opens the named file.
	//
	// When Open returns an error, it should be of type *PathError
	// with the Op field set to "open", the Path field set to name,
	// and the Err field describing the problem.
	//
	// Open should reject attempts to open names that do not satisfy
	// ValidPath(name), returning a *PathError with Err set to
	// ErrInvalid or ErrNotExist.
	Open(name string) (File, error)
}
```

We could use `fs.ReadDirFS`
```go
// ReadDirFS is the interface implemented by a file system
// that provides an optimized implementation of ReadDir.
type ReadDirFS interface {
	FS

	// ReadDir reads the named directory
	// and returns a list of directory entries sorted by filename.
	ReadDir(name string) ([]DirEntry, error)
}
```
to be able to traverse entries, but I'm not sure it would be more convenient to use instead of a list of named bytes.

run.go
akaswenwilk: hmm, I'm not so sure how simple this would be.  From what I can see, if someone wants to run through FeatureContents only, and they leave the Paths option blank, the current behavior will force them to use a features directory by default.  Either they'll have their features already in a features directory, or they won't, in which case there could be a failure due to non-existent directory, no?  I think it's maybe just simpler to say, either you can use feature contents, or features, but not both.  If someone is already parsing files and passing them in, I can't imagine it's much more challenging to also include from all features as opposed to from only a few.  what do you think?

vearutop: Hm, this is interesting point. 

> If someone is already parsing files and passing them in, I can't imagine it's much more challenging to also include from all features as opposed to from only a few.

Another case when this feature could be useful would be sourcing gherkin files from a remote storage (e.g. via HTTP) or generating/templating them by code. In such situation implementing additional file preloader might be a burden.

If we disable that default in case of feature contents presence it may be perceived as unexpected behavior.
```go
	if len(opt.Paths) == 0 && len(opt.Features) == 0 {
		inf, err := os.Stat("features")
		if err == nil && inf.IsDir() {
			opt.Paths = []string{"features"}
		}
	}
```

If we ignore `Paths` (make feature contents and paths mutually exclusive) this can also be perceived as unexpected behavior (that needs to be documented or maybe even have a check and explicit failure to not mislead user).
And implementing preloader can be a minor annoyance for a user in case they want both types of features.

From these options I'm leaning to disabling the default as it seems less disruptive overall.

akaswenwilk: hmm, ok, then how about something like this?
```
	if len(opt.Paths) == 0 && len(opt.FeatureContents) == 0 {
		inf, err := os.Stat("features")
		if err == nil && inf.IsDir() {
			opt.Paths = []string{"features"}
		}
	}

	if len(opt.FeatureContents) > 0 {
		features, err := parser.ParseFromBytes(opt.Tags, opt.FeatureContents)
		if err != nil {
			fmt.Fprintln(os.Stderr, err)
			return exitOptionError
		}
		runner.features = append(runner.features, features...)
	}

	if len(opt.Paths) > 0 {
		features, err := parser.ParseFeatures(opt.Tags, opt.Paths)
		if err != nil {
			fmt.Fprintln(os.Stderr, err)
			return exitOptionError
		}
		runner.features = append(runner.features, features...)
	}
```

akaswenwilk: Sounds good!

vearutop (approved):

PR #509, opened 2022-10-26
Implement unambiguous keywords

### LINKED ISSUE #479: Implement unambiguous keywords
As described in https://github.com/cucumber/common/issues/768

otrava7:
### 🤔 What's changed?

- Introduced public functions `Given()`, `When()`, and `Then()` to `test_context.go`
- Functions work identically to `Step()` but only match a feature step when their specific keyword is used

Resolves #479 

### 📋 Checklist:

- [X] I've changed the behaviour of the code
  - [X] I have added/updated tests to cover my changes.
- [X] My change requires a change to the documentation.
  - [X] I have updated the documentation accordingly.
- [X] Users should know about my change

----


otrava7:
Linking https://github.com/cucumber/common/issues/768

otrava7:
Updated tests, should increase diff coverage to 100%

otrava7:
Thanks for the review! I did see that there was a change in strategy for dependencies, but I was having trouble importing those dependencies from those new repos. The primary reason why I switched to the common repo to begin with is that this PR needed those dependencies to be updated but `go mod` would not allow me as it was saying that the new version declares its name differently than the path. I believe this would occur after this change:
https://github.com/cucumber/messages-go/commit/79a9301661833b46c7a7c1cee5b0d82236b09238

I was facing the same issues importing cucumber/messages repo, due to the same issue, `go mod` was saying the following:
```
module declares its path as: github.com/cucumber/common/messages/go/v19
	        but was required as: github.com/cucumber/messages/go/v19
```
Is this something that can be fixed through a different way of importing that I'm not aware of? Or do we need to change how these modules declare their path?

Either way I'm happy to open new PR's to fix this and then rebase.

otrava7:
Oh, I think I might be able to fix that with a replace directive in the `go.mod`. Just looking for confirmation that is how we want to approach that.

otrava7:
It looks like progress on getting the go part of https://github.com/cucumber/messages and https://github.com/cucumber/gherkin usable has somewhat stalled. Would it be helpful to upgrade these to the usable common monorepo for now, and have another PR to migrate to the newer repos once they become usable?

otrava7:
Rebased branch to cover keyword functions only

vearutop (approved):

cycle-labs-brian (approved):

dumpsterfireproject (approved):

Issue #513, opened 2022-12-27
cancel context at the end of the scenario


draganm:
### 🤔 What's the problem you're trying to solve?

When setting up mock servers running in separate go-routines, it would be useful when the context passed to the scenario `Before` hooks would be cancellable and be cancelled at the end of the scenario.

### ✨ What's your proposed solution?

Wrap the scenario context with `context.WithCancel()` before `Before` hooks are run and call the `cancel()` after last `After()` hook is being called.

### ⛏ Have you considered any alternatives or workarounds?

currently I achieve this with:

```go
func InitializeScenario(ctx *godog.ScenarioContext) {
	var cancel context.CancelFunc

	ctx.Before(func(ctx context.Context, sc *godog.Scenario) (context.Context, error) {
		ctx, cancel = context.WithCancel(ctx)
		return ctx, nil

	})

	ctx.After(func(ctx context.Context, sc *godog.Scenario, err error) (context.Context, error) {
		cancel()
		return ctx, nil
	})

}
```
it's not very elegant and gets repeated every time I use godog

### 📚 Any additional context?
My expectation was that this happens automatically, but only after trawling through the source code and running a couple of experiments it turned out it is not the case.

----

*This text was originally generated from a [template](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/about-issue-and-pull-request-templates), then edited by hand. [You can modify the template here.](https://github.com/cucumber/.github/edit/main/.github/ISSUE_TEMPLATE/feature_request.md)*

vearutop:
Sounds like a reasonable idea, seems this can be implemented here https://github.com/cucumber/godog/blob/202882807b9e70ddc3f6def103edeb3313468958/suite.go#L424-L428 with deferred cancellation in the same function.

I'd be happy to review the PR. 👍

vearutop:
Done in #514.

PR #514, opened 2022-12-27
cancel context for each scenario

draganm:
### 🤔 What's changed?

Each Scenario gets a cancellable context that is cancelled at the end of the scenario.

Cancelled context can be used to shut down mock services that are run in the scope of one Scenario (see #513)

### 📋 Checklist:

----

### DIFF
--- a/suite.go
+++ b/suite.go
@@ -447,9 +447,7 @@ func (s *suite) runPickle(pickle *messages.Pickle) (err error) {
 		ctx = context.Background()
 	}
 
-	var cancel context.CancelFunc
-
-	ctx, cancel = context.WithCancel(ctx)
+	ctx, cancel := context.WithCancel(ctx)
 
 	defer cancel()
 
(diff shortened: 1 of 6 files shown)


draganm:
Not sure how to end-to-end test this though.

mirogta (commented):
This looks great. Would you mind adding a simple example code as well as part of this PR? I would prefer if each use case (like your new one, however it looks cool) has an appropriate example to follow otherwise some less experienced developers might be a bit lost in how to use it.

vearutop:
I think this is a good idea, although it potentially can break existing scenarios that depend (implicitly or explicitly) on background tasks after the scenario has ended already. 
In such (rare?) cases, context cancellation can be canceled 🙃 with help of a simple wrapper.

suite.go
```diff
@@ -447,6 +447,12 @@ func (s *suite) runPickle(pickle *messages.Pickle) (err error) {
 		ctx = context.Background()
 	}
 
+	var cancel context.CancelFunc
+
+	ctx, cancel = context.WithCancel(ctx)
```
vearutop, line 452:
```suggestion
	ctx, cancel := context.WithCancel(ctx)
```

vearutop:
Actually, I think this is good to merge, and I'll add a test in a follow-up commit.

PR #538, opened 2023-02-14
docs(*): correct example

otrava7:
### 🤔 What's changed?

- Correct docs example to use right keyword

- :bank: Refactoring/debt/DX (improvement to code design, tooling, documentation etc. without changing behaviour)

----

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,8 +158,8 @@ func thereShouldBeRemaining(arg1 int) error {
 }
 
 func InitializeScenario(ctx *godog.ScenarioContext) {
-        ctx.Step(`^I eat (\d+)$`, iEat)
         ctx.Step(`^there are (\d+) godogs$`, thereAreGodogs)
+        ctx.Step(`^I eat (\d+)$`, iEat)
         ctx.Step(`^there should be (\d+) remaining$`, thereShouldBeRemaining)
 }
 ```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -167,8 +167,8 @@ func InitializeScenario(ctx *godog.ScenarioContext) {
 Alternatively, you can also specify the keyword (Given, When, Then...) when creating the step definitions:
 ``` go
 func InitializeScenario(ctx *godog.ScenarioContext) {
-        ctx.Given(`^I eat (\d+)$`, iEat)
```
mirogta, line 170: Thank you for your contribution!

I can see that you have reordered the Given and When calls. Actually, the logical order of calls and setting them up was correct - Given, When, Then - so the new order When, Given, Then actually is counter-intuitive.

Can you please revert your change and instead reorder the funcs above? Thx

mirogta (changes requested):

README.md
otrava7: Sure, I will put the Given statement first. The reason I made the change is because in the feature file "Given" is linked to "there are 12 godogs" and "When" is linked to "I eat 5"

mirogta (approved):
LGTM - thanks!

Issue #100, opened 2017-09-14
Add support for testing.T in steps


maxmeyer:
# Summary

I built a library - a very basic version of what https://github.com/cucumber/aruba does for ruby - which is used with `go test`. 

# Details 

Using the standard library I have `testing.T` available to pass to the helpers - e.g. to mark the method via `t.Helper()`. Now I would love to use it with `godog`.

# Question

What would you recommend to do next? I think about adding some kind of a compatibility layer to the library. Or can we have `testing.T` on steps in "godog" as well?

Sorry the library is not (yet) public.

l3pp4rd:
Hi, I think testing.T is something worth considering, but I'm afraid such methods as t.Fatal are something which godog cannot handle and they will be exposed. If we can make t.Fatal or t.Error captured by godog in order to print properly the error or stop suite execution after printing it after t.Fatal then it would be possible to integrate it.

maxmeyer:
Not sure, but somehow to test seems to manage it as well.

creack:
This would be really nice indeed. A solution would be to do something similar as what the testify lib does and have a subset of testing.T like

```
// TestingT is an interface wrapper around *testing.T
type TestingT interface {
        Errorf(format string, args ...interface{})
        FailNow()
}
```

tylerauerbeck:
Sorry to revive an old issue here. Is this issue still valid? Or is there a way to use testing.T now?

lonnblad:
Hi @tylerauerbeck, no need to be sorry and I think the issue is still valid.

I would definitely like to look at doing something like @creack is suggesting, but we are a pretty small group that is currently maintaining and developing godog and we haven't prioritised this yet. If you are interested in doing some work, I would be happy to collaborate! 😄

lonnblad:
Hi, if you have time, please check out the suggestion in #360 and comment there if you have would like to see changes to the draft.

aaronc:
Without support for this I'm having a hard time figuring out how to integrate something like `gomock` where the `Controller` needs to be initialized with something resembling `testing.T`: https://pkg.go.dev/github.com/golang/mock@v1.6.0/gomock#NewController

bryanchriswhite:
[...]

PR #570, opened 2023-08-14
Provide access to the testing.T log function in the context

mrsheepuk:
### 🤔 What's changed?

Simply adding an interface over testing.T's log and logf functions to the context, allowing logging of information contextually. Vaguely related to #535 and #100 as, if this pattern is OK, it could be expanded to provide the whole of testing.T.

Example usage:
```
func log(ctx context.Context, msg string, args ...interface{}) {
	if t := godog.GetTestLogger(ctx); t != nil {
		t.Logf(msg, args...)
	} else {
		fmt.Println(fmt.Printf(msg, args...))
	}
}
```

We have longer-running asynchronous tests which, when run, we want to provide contextual feedback to the user on during the execution. By using the sub-test's testing.T, these log messages appear in a sane place in the formatted output.

Opening this early for feedback - if it is conceptually sane, I'll work this PR up to a better state and add docs - didn't want to bother documenting it in detail if it's not going to be acceptable tho!

### 📋 Checklist:

### DIFF
--- a/suite.go
+++ b/suite.go
@@ -522,6 +522,7 @@ func (s *suite) runPickle(pickle *messages.Pickle) (err error) {
 	if s.testingT != nil {
 		// Running scenario as a subtest.
 		s.testingT.Run(pickle.Name, func(t *testing.T) {
+			ctx = setContextLogger(ctx, t)
 			ctx, err = s.runSteps(ctx, pickle, pickle.Steps)
 			if s.shouldFail(err) {
 				t.Errorf("%+v", err)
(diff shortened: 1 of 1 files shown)


tigh-latte:
I like this idea, but, the use of `if` and `else` to change between loggers makes me think we could improve the API.

Given that `log/slog` now has a few functions which take `context.Context` as their first param, perhaps we could follow suit and expose a few log functions in `godog`. So your example would be something like:

```go
func log(ctx context.Context, msg string, args ...interface{}) {
	godog.Logf(ctx, msg, args...)
}
```

Under the hood, that/(those?) function would check if we have a logger in the `context.Context`:
```go
func Logf(ctx context.Context, msg string, args ...any) {
	if logger := getLogger(ctx); logger != nil {
		logger.Logf(msg, args...)
	} else {
		fmt.Println(msg+"\n", args...)
	}
}
```

`godog` would then expose every function in `testing.T` that we are comfortable with exposing.

mrsheepuk:
Ooooh, yes, I see what you mean @tigh-latte - I'll have a further play and see if I can come up with a pattern that also solves the testify/assert problem (as I also hit that) without breaking the fundamental way of working...

PR #582, opened 2023-10-23
updated base formatter to set a scenario as passed unless there exist…

### LINKED ISSUE #581: Why Getting total scenarios as 6 while I am having total 5 scenarios
Screenshot 2023-10-23 at 11 11 26 AM

I am running test as go test -v

getting total 5 scenrioes in which 2 passed and 4 fails.

**Expected result := 5 (1 passed, 4 failed)**

roskee:
### 🤔 What's changed?
I have updated the step evaluation in the base formatter to not set a scenario as passed just because some step has passed. A scenario, after this change, will only be passed only if no steps exist that are `failed`, `pending` or `undefined` in that scenario.

Issue #581 was created because of this. I was able to reproduce it easily. All I had to do was update the [api test feature] like the following (only updated the `method not allowed` to `method allowed`
```
# file: version.feature
Feature: get version
  In order to know godog version
  As an API user
  I need to be able to request version

  Scenario: does not allow POST method
    When I send "POST" request to "/version"
    Then the response code should be 405
    And the response should match json:
      """
      {
        "error": "Method allowed"
      }
      """

  Scenario: should get version number
    When I send "GET" request to "/version"
    Then the response code should be 200
    And the response should match json:
      """
      {
        "version": "v0.0.0-dev"
      }
      """

```
The summary for the above change is as follows
image

As it can be seen from the picture, the total scenarios (2) does not much the sum of passed (2) and failed (1) scenarios. This is due to the steps overriding the scenario status. That is, if a failed step is evaluated first, it will set the scenario as failed. but the next passing step will override that state and make it as passed. But when the summary is prepared the number of failed scenarios is evaluated as the number of failed steps (since a failed step will always lead to a failed scenario and other dependent steps are skipped, this logic works). This leads to the incorrect sum in the above result.

I am not sure if this issue is only on the base formatter implementation. If there are any other implementations of the formatter I should look into, please let me know :pray: 

### 📋 Checklist:

----

### DIFF
--- a/internal/formatters/fmt_base.go
+++ b/internal/formatters/fmt_base.go
@@ -106,7 +106,6 @@ func (f *Base) Summary() {
 
 			switch sr.Status {
 			case passed:
-				prStatus = passed
 				passedSt++
 			case failed:
 				prStatus = failed
@@ -114,10 +113,8 @@ func (f *Base) Summary() {
 			case skipped:
 				skippedSt++
 			case undefined:
-				prStatus = undefined
 				undefinedSt++
 			case pending:
-				prStatus = pending
 				pendingSt++
 			}
 		}
--- a/internal/formatters/fmt_base.go
+++ b/internal/formatters/fmt_base.go
@@ -113,8 +113,10 @@ func (f *Base) Summary() {
 			case skipped:
 				skippedSt++
 			case undefined:
+				prStatus = undefined
 				undefinedSt++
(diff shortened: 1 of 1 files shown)


roskee:
@vearutop Is there anything I should update before you can approve this?

vearutop:
@roskee thank you for fixing this!

PR #605, opened 2024-01-26
Fixes #604 - remove line overwriting for scenario outlines in cucumber formatter

### LINKED ISSUE #604: Unexpected step line values in cucumber report generated by Godog for scenario outlines
### Discussed in https://github.com/cucumber/godog/discussions/603

<div type='discussions-op-text'>

<sup>Originally posted by **katyaaronov** January 24, 2024</sup>
The below code from cucumber/godog/fmt_cucumber.go overwrites the line property values of steps in cucumber reports (for Scenario Outlines only, regular Scenarios reports look good).

As a result, the step line property corresponds to the line in the Examples section rather than to the actual location of step within the feature file.

Due to the above, external tools that process the reports and use the line property for sorting the steps, display the steps out of order.

Other officially supported cucumber implementations (java, js, python, etc.) are generating reports correctly for Scenario Outline - line property corresponds to the actual location of step within the file rather than the line in Examples section.

Please advice.

Regards, Katya.

```
func (f *Cuke) buildCukeStep(pickle *messages.Pickle, stepResult models.PickleStepResult) (cukeStep cukeStep) {

    ...

    line := step.Location.Line
    if len(pickle.AstNodeIds) == 2 {
        _, row := feature.FindExample(pickle.AstNodeIds[1])
        line = row.Location.Line
    }

    ...

    cukeStep.Line = int(line)

    ...
}
```

glibas:
### 🤔 What's changed?
Removing line overwrite from cucumber formatter 
```
if len(pickle.AstNodeIds) == 2 {
        _, row := feature.FindExample(pickle.AstNodeIds[1])
        line = row.Location.Line
    }
```

The above code from cucumber/godog/fmt_cucumber.go overwrites the line property values of steps in cucumber reports (for Scenario Outlines only, regular Scenarios reports look good).

As a result, the step line property corresponds to the line in the Examples section rather than to the actual location of step within the feature file.

Due to the above, external tools that process the reports and use the line property for sorting the steps, display the steps out of order.

Other officially supported cucumber implementations (java, js, python, etc.) are generating reports correctly for Scenario Outline - line property corresponds to the actual location of step within the file rather than the line in Examples section.

### 📋 Checklist:

  - [x] I have updated tests to cover my changes.

----

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,9 @@ This project adheres to [Semantic Versioning](http://semver.org).
 This document is formatted according to the principles of [Keep A CHANGELOG](http://keepachangelog.com).
 
 ## Unreleased
-
+### Fixed
+- Remove line overwriting for scenario outlines in cucumber formatter ([605](https://github.com/cucumber/godog/pull/605) - [glibas](https://github.com/glibas))
+  
 ## [v0.13.0]
 ### Added
 - Support for reading feature files from an `fs.FS` ([550](https://github.com/cucumber/godog/pull/550) - [tigh-latte](https://github.com/tigh-latte))
--- a/features/formatter/cucumber.feature
+++ b/features/formatter/cucumber.feature
@@ -323,7 +323,7 @@ Feature: cucumber json formatter
                 {
                   "keyword": "Given ",
                   "name": "passing step",
-                  "line": 11,
+                  "line": 7,
                   "match": {
                     "location": "suite_context.go:64"
                   },
@@ -345,7 +345,7 @@ Feature: cucumber json formatter
                 {
                   "keyword": "Given ",
                   "name": "failing step",
-                  "line": 12,
+                  "line": 7,
                   "match": {
                     "location": "suite_context.go:47"
                   },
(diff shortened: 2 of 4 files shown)


glibas:
Hi @vearutop could you please take a look

roskee:
@glibas could you perhaps share an example to reproduce the problem so that we can test it as well. And also link the original issue #604 so that it will be closed once this pull request is done. Thanks :slightly_smiling_face:

glibas:
Hi @roskee ,
The issue is well illustrated in this test cucumber.json output [scenario_outline](https://github.com/cucumber/godog/pull/605/files#diff-1a2b56b32c938ea2538103d31b18d669e8ae96f08beacd0b225b0697899e73ac). The line of the step is being overwritten with the line of the example in Examples table, so all the steps within one scenario outline instance have the same line number causing
 report generators which are using cucumber.json output to display steps in random order, since they rely on the line number.

glibas:
Hi @roskee, also as an example can be used this scenario [Scenario: Support of Feature Plus Scenario Outline With Steps](https://github.com/cucumber/godog/blob/5a5631a758b53795d388b96e25126537396506ea/features/formatter/cucumber.feature#L286C3-L286C64) which can be executed before the change and after the change to observe changes in lines numbering

roskee:
Sorry for being late. I was very busy. I will take a look and let you know my thoughts.

glibas:
No problem, thank you. In essence the change aligns godog cucumber json reporting for scenario outlines with other cucumber implementation so hopefully should be straight forward.

roskee (approved):
I think this is the right way. I am not sure why step lines were being overwritten in the first place :man_shrugging:

glibas:
Thank you @roskee for confirming!

Issue #617, opened 2024-04-29
Support for attachments / embeddings as per Cucumber JVM and Behave


Johnlon:
### 🤔 What's the problem you're trying to solve?

In cucumber jvm the Scenario class provides a means to attach(bytes, mediator, name) but we can't see the same  capability in godog.
How do we add typed attachments to the output report file?

### ✨ What's your proposed solution?

Add the necessary API and docs

### ⛏ Have you considered any alternatives or workarounds?

None known

Johnlon:
Hi take a look at cucumber JVM JSON report output.....
It has this at the step level ...

"embeddings" [ 
 {
  "data" : "string of base 64 data",
  "mine_type" : "string",
  "name": "string"
 }
];

In the JVM api see the api like .. Scenario.attach(name, mimetype, data)

The newer "messages" format in Cucumber JVM has the EmbedEvent which also represents this data.

Additionally the new Cucumber HTML formatter (Cucumber React) consumes these embedding messages and displays them correctly.

---

Re the attach API .. 

The same kind of thing is in the main branch of Behave but because Behave looks like it's dying (ie it doesn't make prod releases anymore, 7 years) then I guess folk build their own dist of Behave containing that feature.

Johnlon:
I think there is a deeper problem and that is that the cucumber/messages repo isn't being kept in sync with things like CucumberJVM.   Cucumber JVM defines it's own message types I think and something similar in CucumberReact so if godog is relying on the Cucumber Messages repo for a standard then I think that's a problem.

Johnlon:
Ob might be thibnking of this ... https://github.com/cucumber/cucumber-json-schema/issues/6

vearutop:
I think this https://github.com/cucumber/messages ([json schema](https://github.com/cucumber/messages/tree/main/jsonschema)) is the current main reporting format (and it is used in godog internally).

What I'm interested in is a full example file that exhibits variable attachments in different places (also if possible the case of multiple attachments per step result), because I'm not super fluent in these semantics and where attachments should be applicable and how the relation is built.

vearutop:
From the implementation perspective I think we can introduce something like `ctx := godog.AddAttachment(ctx, attachment)` and then catch that information between the steps.

Johnlon:
> From the implementation perspective I think we can introduce something like `ctx := godog.AddAttachment(ctx, attachment)` and then catch that information between the steps. 

Steps may add multiple attachments.
[...]

PR #623, opened 2024-05-08
added support for Attachments (aka Embedddings)

Johnlon:
### 🤔 What's changed?

The cucumber and events formatters support the inclusion of Attachments.

The proposed API is demonstrated in the tests as  ...

    func stepWithAttachment(ctx context.Context) (context.Context, error) {
	    ctxOut := godog.Attach(ctx,
		godog.Attachment{Body: []byte("TheData1"), FileName: "TheFilename1", MediaType: "text/plain"},
		godog.Attachment{Body: []byte("TheData2"), FileName: "TheFilename2", MediaType: "text/plain"},
	    )

	    return ctxOut, nil
    }

https://github.com/cucumber/godog/issues/617

Want support for Attachments from steps.
This PR proposes to use the context.Context object as a means to attach content from the step.

This change involves minimal changes to patterns of use of godog API's.

I added tests for the cuke and events formats but the coverage isn't being recorded even on some lines I didn't modify.

fmt_output_test.go now has a normalise() fn that eliminates some extraneous diffs in the actual/expected files such as line endings and also line numbers so that the tests are easier to maintain and cross platform,

### 📋 Checklist:

### DIFF
--- a/internal/formatters/fmt_cucumber.go
+++ b/internal/formatters/fmt_cucumber.go
@@ -12,6 +12,7 @@ package formatters
 */
 
 import (
+	"encoding/base64"
 	"encoding/json"
 	"fmt"
 	"io"
@@ -139,14 +140,21 @@ type cukeMatch struct {
 	Location string `json:"location"`
 }
 
+type cukeEmbedding struct {
+	Name     string `json:"name"`
+	MimeType string `json:"mime_type"`
+	Data     string `json:"data"`
+}
+
 type cukeStep struct {
-	Keyword   string              `json:"keyword"`
-	Name      string              `json:"name"`
-	Line      int                 `json:"line"`
-	Docstring *cukeDocstring      `json:"doc_string,omitempty"`
-	Match     cukeMatch           `json:"match"`
-	Result    cukeResult          `json:"result"`
-	DataTable []*cukeDataTableRow `json:"rows,omitempty"`
+	Keyword    string              `json:"keyword"`
+	Name       string              `json:"name"`
+	Line       int                 `json:"line"`
+	Docstring  *cukeDocstring      `json:"doc_string,omitempty"`
+	Match      cukeMatch           `json:"match"`
+	Result     cukeResult          `json:"result"`
+	DataTable  []*cukeDataTableRow `json:"rows,omitempty"`
+	Embeddings []*cukeEmbedding    `json:"embeddings,omitempty"`
 }
 
 type cukeDataTableRow struct {
@@ -294,6 +302,21 @@ func (f *Cuke) buildCukeStep(pickle *messages.Pickle, stepResult models.PickleSt
 		cukeStep.Match.Location = fmt.Sprintf("%s:%d", pickle.Uri, step.Location.Line)
 	}
 
+	if stepResult.Attachments != nil {
+		attachments := []*cukeEmbedding{}
+
+		for _, a := range stepResult.Attachments {
+			attachments = append(attachments, &cukeEmbedding{
+				Name:     a.Name,
+				Data:     base64.RawStdEncoding.EncodeToString(a.Data),
+				MimeType: a.MimeType,
+			})
+		}
+
+		if len(attachments) > 0 {
+			cukeStep.Embeddings = attachments
+		}
+	}
 	return cukeStep
 }
 
--- a/internal/models/results.go
+++ b/internal/models/results.go
@@ -18,6 +18,13 @@ type PickleResult struct {
 	StartedAt time.Time
 }
 
+// PickleAttachment ...
+type PickleAttachment struct {
+	Name     string
+	MimeType string
+	Data     []byte
+}
+
 // PickleStepResult ...
 type PickleStepResult struct {
 	Status     StepResultStatus
@@ -28,6 +35,8 @@ type PickleStepResult struct {
 	PickleStepID string
 
 	Def *StepDefinition
+
+	Attachments []*PickleAttachment
 }
 
 // NewStepResult ...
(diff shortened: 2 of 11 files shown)


mattwynne:
@Johnlon thanks for this, it looks great!

Are you able to take a look at why the build is failing? It seems to be just some noise from `go fmt`, maybe your editor has added some whitespace or something?

mattwynne:
@Johnlon it would help the other maintainers if you could add an entry to the changelog as part of this PR. You can use [this tool](https://github.com/cucumber/changelog) if you like, or just copy and adapt other existing entries.

Johnlon:
@mattwynne thanks - will do.
Still confused by the coverage problems.
Any ideas?
Cheers John

Johnlon:
> @Johnlon thanks for this, it looks great!
> 
> Are you able to take a look at why the build is failing? It seems to be just some noise from `go fmt`, maybe your editor has added some whitespace or something?

At go fmt sorry .
Hmm will see if I can reverse??

mattwynne:
Hey @Johnlon no worries, looks like the build on main is green which is the main thing I was worried about, so all good.

As I said in the other thread, we favour [Ship/Show/Ask](https://martinfowler.com/articles/ship-show-ask.html) in the Cucumber org, so feel free to merge your own PRs if you feel you've had enough feedback. Try and do it on purpose though! 😆

mattwynne:
I really appreciate the work you've put into this, thanks!

internal/models/results.go
```diff
@@ -18,6 +18,13 @@ type PickleResult struct {
 	StartedAt time.Time
 }
 
+// PickleAttachment ...
+type PickleAttachment struct {
```
vearutop, line 22: Is this type needed, or can we use `godog.Attachment` instead?

internal/models/results.go
```diff
@@ -28,13 +35,16 @@ type PickleStepResult struct {
 	PickleStepID string
 
 	Def *StepDefinition
+
+	Attachments []*PickleAttachment
```
vearutop, line 39: I'd change this to slice of values if there is no specific reason for a slice of pointers.

suite.go
```diff
@@ -68,6 +68,42 @@ type suite struct {
...
 
+type Attachment struct {
+	Body      []byte
+	FileName  string
+	MediaType string
+}
+
+type attachmentKey struct{}
+
+func Attach(ctx context.Context, attachments ...Attachment) context.Context {
```
vearutop, line 79: The API of `Attach`/`Attachments` seems reasonable, but maybe needs a bit of documentation (as a function comment) on the expected behavior. One may think that

```go
ctx = godog.Attach(ctx, a1)
// ..... things happening here
ctx = godog.Attach(ctx, a2)
```
would add both `a1` and `a2` as attachments, though instead they would need to use something like

```go
ctx = godog.Attach(ctx, a1)
// ..... things happening here
ctx = godog.Attach(ctx, append(godog.Attachments(ctx), a2)
```

vearutop (commented):
Thank you! Sorry, I'm late to the party, but probably we can consider/apply a few changes on top of this PR (with a new PR) before we tag a new version (to avoid backwards compatibility concerns).

Johnlon:
No prob happy to help
How do releases work?
Thanks

mattwynne:
First things first, before thinking about a release, if you can make another PR with the changes @vearutop has suggested, then we'll have the code in the best shape we can get it.

internal/formatters/fmt_cucumber.go
```diff
@@ -139,14 +140,21 @@ type cukeMatch struct {
...
-	Result    cukeResult          `json:"result"`
-	DataTable []*cukeDataTableRow `json:"rows,omitempty"`
+	Keyword    string              `json:"keyword"`
+	Name       string              `json:"name"`
+	Line       int                 `json:"line"`
+	Docstring  *cukeDocstring      `json:"doc_string,omitempty"`
+	Match      cukeMatch           `json:"match"`
+	Result     cukeResult          `json:"result"`
+	DataTable  []*cukeDataTableRow `json:"rows,omitempty"`
+	Embeddings []*cukeEmbedding    `json:"embeddings,omitempty"`
```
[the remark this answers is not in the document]
Johnlon: Changed - the impact will be more copying of the attachment values so a slight hit in terms of perf and mem, but probably irrelvant.

suite.go
```diff
@@ -68,6 +68,42 @@ type suite struct {
 	afterScenarioHandlers  []AfterScenarioHook
 }
 
+type Attachment struct {
```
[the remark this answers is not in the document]
Johnlon: Disagree on this one.

This aliasing isn't into user space and in any case except for GherkinDocument the alising is only the pickles.

Also using mesaages.Attachment would mean taking an internal message class and exposing the end user to it, which will may increate coupling. Also there is no reduction in data mapping as the cuke report has it's own data structures to populate and because there is no other usage of those message classes in godog as there is no support for the Messaages format yet (I have a separate PR coming for that).

internal/models/results.go
Johnlon: See my previous comments messages.Attachment should remain and internal class.

suite.go
Johnlon: I've improved the impl so its convenient in the situation mentioned.

This now works ..

    ctx = godog.Attach(ctx, a1)
    // ..... things happening here  
    ctx = godog.Attach(ctx, a2)

Johnlon:
@vearutop  Hi - Take a look at my responses to your suggestions.
PR raised here : https://github.com/cucumber/godog/pull/628

Issue #633, opened 2024-06-21
runAfterStepHooks should allow the user to replace an error with a nil or another error


Johnlon:
### 🤔 What's the problem you're trying to solve?

I need the ability to installa hook that either knocks out an error returned from a step, or completely replaced the error returned from a step.

The current runAfterStepHooks function does not make this possible as it has a guard expression that prevents me knocking out a current error, and a separate guard expression that prevents me substituting a different error (does an append instead).

### ✨ What's your proposed solution?

I can't see a backwards compat solution so I will raise a PR adding a "PostStepHook" and runPostStepHooks()  which is a more open api than the existing PostStepHook. 

### ⛏ Have you considered any alternatives or workarounds?

Alternatively, we can break backwards compatibility -  we could replace the existing runAfterStepHooks  with the impl I have in runPostStepHooks in my PR.

The two impls are below for your convenience....

```go
func (s *suite) runPostStepHooks(ctx context.Context, step *Step, status StepResultStatus, err error) (context.Context, StepResultStatus, error) {
	for _, f := range s.postStepHandlers {
		ctx, status, err = f(ctx, step, status, err)
	}

	return ctx, status, err
}

func (s *suite) runAfterStepHooks(ctx context.Context, step *Step, status StepResultStatus, err error) (context.Context, error) {
	for _, f := range s.afterStepHandlers {
		hctx, herr := f(ctx, step, status, err)

		// Adding hook error to resulting error without breaking hooks loop.
		if herr != nil {
			if err == nil {
				err = herr
			} else {
				err = fmt.Errorf("%v, %w", herr, err)
			}
		}

		if hctx != nil {
			ctx = hctx
		}
	}

	return ctx, err
}
```

### 📚 Any additional context?

It is unfortunate that the existing AfterStep hook make the assumptions it does in runAfterStepHooks.

"AfterStep" doco says 
  // It may be convenient to return a **different** kind of error
  // in order to print more state details which may help  
... but the current impl doesn't allow me to return a "**different"** error, just an **"additional**" error.

PR #634, opened 2024-06-21
PostStepHook is what I think AfterStepHook ought to have been - https://github.com/cucumber/godog/issues/633

Johnlon:
# See https://github.com/cucumber/godog/issues/633

### 🤔 What's changed?

https://github.com/cucumber/godog/issues/633 covers the proposal.

AfterStepHook doesn't allow full manipulation of the test outcome, so added PostStepHook to compensate without breaking compat.

I need full control over the test outcome from a hook and the AfterStepHook does not permit that.

To be honest I'd rather just introduct a breaking change and replace the logic of AfterStep with my proposed PostStep.

### 📋 Checklist:

- [x ] I've changed the behaviour of the code
- [ x] My change requires a change to the documentation.
- [ x] Users should know about my change

I've added an _example but not figured out the testing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ This document is formatted according to the principles of [Keep A CHANGELOG](htt
 
 ## Unreleased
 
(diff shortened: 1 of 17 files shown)


Johnlon:
Not going to progress

PR #642, opened 2024-09-29
docs: add duplicate steps example and "Motivation" README

noodnik2:
### 🤔 What's changed?

Added an example (i.e., `_examples/dupsteps`) to illustrate the "problem" we're facing and essentially the request - 
either for suggestions for addressing the problem, or to enable us to (preferably) tie a step function to a scenario (as opposed to the "global" association it appears to be now), or fail (or at least issue a warning) when more than one step function could be associated with a given step's text (as opposed to silently forgetting the overridden step function as it appears to do now).

The motivation is described in the accompanying README file.

No changes at this point, just a clarification, tickler for discussion, and initiative to help me understand better how to use the tool.  It _might_ lead to a change that I'd be happy to take on and submit in a future PR, depending upon the outcome of the discussion.

Yes, everything written above and in the included `README` file. 

### 📋 Checklist:

----


_examples/dupsteps/dupsteps_test.go
```diff
@@ -0,0 +1,113 @@
...
+		Given I accidentally poured concrete down my drain and clogged the sewer line
+		Then I fixed it
+		Then I can once again use my sink
+`
+
+	suite := godog.TestSuite{
+		Name: t.Name(),
+		ScenarioInitializer: func(context *godog.ScenarioContext) {
+			// NOTE: loading implementations of steps for different scenarios here
+			(&cloggedDrain{}).addCloggedDrainSteps(context)
```
noodnik2, line 32: The "clogged drain" scenario (above) works fine until the "flat tire" scenario (below) is added (e.g., in a subsequent commit).

The _"I fixed it"_ step from the newly added `flatTire` scenario silently overrides the matching step in the previously defined step from the `cloggedDrain` scenario.  The writer of the clogged drain scenario will have to investigate why his/her scenario suddenly stops working.

We'd like the tool to support a distributed development workflow where developers can work on scenarios independently, without having to take into account (and reuse) all pre-existing, matching steps.

_examples/dupsteps/dupsteps_test.go
```diff
@@ -0,0 +1,113 @@
...
+
+// Implementation of the steps for the "Flat Tire" scenario
+
+type flatTire struct {
+	tireIsFlat bool
+}
+
+func (ft *flatTire) addFlatTireSteps(ctx *godog.ScenarioContext) {
+	ctx.Step(`^I ran over a nail and got a flat tire$`, ft.gotFlatTire)
+	ctx.Step(`^I fixed it$`, ft.iFixedIt)
```
noodnik2, line 91: NOTE: the text of this new (or the old, pre-existing) step could be something less obviously matching - e.g., `^(.+)\s+fixed\s+(.+)$`, making it harder for the second developer to spot and re-use the step instead of writing their own.

Johnlon:
In the text you discount the idea of running godog multiple times.
However, it's trivial to merge multiple cuke files ones all your invocations have completed.

_examples/dupsteps/cmd/stepchecker/main.go
```diff
@@ -0,0 +1,295 @@
...
+
+	fmt.Println()
+
+	if issuesFound != 0 {
+		log.Printf("%d issue(s) found\n", issuesFound)
+		os.Exit(RC_ISSUES)
+	}
+}
+
+func collectGoSteps(goFilePath string, steps map[string]*StepMatch) error {
```
Johnlon, line 124: I dont know if this static parser will spot the cases where the ScenarioContext has been wrapped.
I've seen (and I use) a scheme where folk register steps via a proxy wrapper of the ScenarioContext, that adds generic behaviours to the steps being registered.

noodnik2: Nice catch! 👍

If you provide an example, I'll work up a test case to fix.

Do you have a fix strategy in mind?

noodnik2:
> _The text from the first para ... "Running the tests for the two Scenarios _separately_ (e.g., using a separate `godog.TestSuite`)..._
> 
> _.... seems contradict the proposed solution_

Hmm, okay I guess that didn't come out so clearly then.  

What I'm trying to say is that by testing Scenarios individually (aka separately), we _can work around_ the problem of ambiguity; but, that _causes_ the problem (for us) of producing a single report covering _all Scenarios_, given that `godog` produces a _separate report for each invocation_ of the encapsulating `godog.TestSuite`.  Looking at the "problem" this way lead me to the proposed `solution` (see folder) of _combining multiple reports into one._

Suggestions welcome for a clearer wording to express this!

PR #645, opened 2024-10-05
fix(formatter): On concurrent execution, execute formatter at end of Scenario

### LINKED ISSUE #643: Running with concurrency shouldn't intermingle the test outputs
### 👓 What did you see?

Example output showing three different test's steps being mixed together
image

### ✅ What did you expect to see?

It would be more readable and easier to debug to hold onto a test's output until it's complete and then write it all at once.

### 📦 Which tool/library version are you using?

_No response_

### 🔬 How could we reproduce it?

Set `Concurrency:  2` (or more) in the config

### 📚 Any additional context?

_No response_

tigh-latte:
### 🤔 What's changed?

Add a new formatter type which only executes when `Flush` is called. Use this formatter when running concurrently.

Then while running concurrently, after the `suite` is copied for a scenario, we overwrite the copy's `Formatter` to use this new `OnFlushWrap` formatter.

At the end of `runPickle`, if we can `Flush`, we do.

Fixes #643 , Scenario output is only printed at end of `runPickle` func when running concurrently.

Any feedback on the general approach to solving this would be good.

### 📋 Checklist:

----


tigh-latte:
one test is broken, `Test_FormatterConcurrencyRun` but on when running `scenario_outline.feature`, keep this as a draft until I work out why.

internal/storage/storage.go
```diff
@@ -233,6 +233,21 @@ func (s *Storage) MustGetPickleStepResultsByPickleID(pickleID string) (psrs []mo
 	return psrs
 }
 
+// MustGetPickleStepResultsByPickleIDUntilStep will retrieve pickle step results by pickle id
+// from 0..stepID for that pickle.
+func (s *Storage) MustGetPickleStepResultsByPickleIDUntilStep(pickleID string, untilStepID string) (psrs []models.PickleStepResult) {
```
tigh-latte, line 238: This only needed to be done for the Pretty formatter, so I wasn't sure if I should just do it after `MustGetPickleStepResultsByPickleID` in `fmt_pretty.go`, or create a function for it.

I've opted to create a function with a pretty bad name, so let me know if yous are happy with this.

vearutop (approved):

vearutop:
Thank you!

Johnlon:
if this is a new formatter please, can we please have fmt-output-test cases in line with the other formattere.

tigh-latte:
@Johnlon it isn't really a new formatter and instead is a wrapper of current formatters. All it does is delay the execution of the wrapped formatter until the end of the scenario, and this delayed execution is tested.

In an additional PR, I could put time into testing that the output of a wrapped formatter is as we would expect, but given the wrapper only provides behavioural changes, testing the behaviour as I've done here and using the output tests already in place is enough.

thomshutt:
@tigh-latte Thanks so much for adding this, I've been able to re-enable concurrency in my tests because of it!

PR #647, opened 2024-10-10
Fix some type checks on the signatures of nested step handlers

Johnlon:
### 🤔 What's changed?

This change fixes some type checks on the signatures and runtime return values of the handlers. 
This change also improves the quality of the error messages produced.

--

History ...

At some point someone changed the return type for "nested steps" handlers from []string to godog.Steps but they forgot to adjust the type checks where the handlers are registered.

They got away with this because the existing type checks were lax and unable to distinguish  []string from godog.Steps.
However,  in a couple of places in the code the value is coerced to godog.Steps and so if someone had actually returned []string then the code would blow up at run time rather than during registration. 

Additionally there were some tests around these types but they also had not been updated and the test was passing for the wrong reason - the particular test expected an error but the cause of the error was an accident caused by the coercion mentioned above rather than a delibarate act of the test.

Bug fix and better error handling.

I'm pretty sure the change is good - comments please.

### 📋 Checklist:

- [ x] I've changed the behaviour of the code
  - [x ] I have added/updated tests to cover my changes.

----

### DIFF
--- a/internal/formatters/fmt_output_test.go
+++ b/internal/formatters/fmt_output_test.go
@@ -179,9 +179,6 @@ func fmtOutputTest(fmtName, testName, featureFilePath string) func(*testing.T) {
 		expected := normalise(string(expectedOutput))
 		actual := normalise(buf.String())
 		assert.Equalf(t, expected, actual, "path: %s", expectOutputPath)
-		if expected != actual {
-			println("diff")
-		}
 	}
 }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ This document is formatted according to the principles of [Keep A CHANGELOG](htt
 
 ## Unreleased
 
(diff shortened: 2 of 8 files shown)


Johnlon:
@mpkorstanje @vearutop - mind taking a look please.

The PR description explains it all.
Pretty sure this is all good but wouldn't mind some other eyes.

internal/formatters/fmt_output_test.go
```diff
@@ -179,9 +179,6 @@ func fmtOutputTest(fmtName, testName, featureFilePath string) func(*testing.T) {
 		expected := normalise(string(expectedOutput))
 		actual := normalise(buf.String())
 		assert.Equalf(t, expected, actual, "path: %s", expectOutputPath)
-		if expected != actual {
```
Johnlon, line 182: I removed some junk

mpkorstanje:
The PR looks reasonable but I don't have enough familiarity to review Go.

vearutop (approved):
LGTM, thanks!

Johnlon:
thanks folk. I have a bunk of other wip I need finish.

PR #659, opened 2024-10-30
DONT MERGE YET - Test renovation - in preparation for fixing the Hook handler bug and adding other features.

Johnlon:
### 🤔 What's changed?

suite_context_test.go is now testing solely using the Public of godog api.
Previously this suite wasn't really testing godog at all, because it was testing a fake godog-like mock up and we were testing .  consequently a bunch of the tests weren't actually doing anything useful.
The most useful tests therefore were the fmt_output_tests which DID go via the real product.

Much of the code changes are a consequence of fixing these issues.

- Fixed broken tests
- Removed duplicates tests
- Removed bogus / useless tests
- Relocated tests
- Added new tests

The public api has an additiive change TestSuite.RunWithResults() that returns the exit code plus a reference to the storage.
This addiitonal public api is really useful for various reasons (eg postprocessing) but importantly its a hook needed to do some of the tests without totally hackery.

Product is a bit too hard to work with and I want to add some features/fixes.
This is only a partial renovation.

### 📋 Checklist:

  - [] I have added an entry to the "Unreleased" section of the [**CHANGELOG**](../blob/main/CHANGELOG.md), linking to this pull request.

----


functional_test.go
```diff
@@ -0,0 +1,1099 @@
...
+	"sort"
+	"strconv"
+	"strings"
+	"testing"
+)
+
+/*
+This file contains some functional tests for the godog library.
+This is the glue code that can be run against features/*feature
+and those feature files are written to use godog features to test godod.
```
noodnik2, line 34: `godod` (sp)

functional_test.go
```diff
@@ -0,0 +1,1099 @@
...
+
+The general pattern is the feature files have top level or "outer" steps with
+that carry mini-features in docstrings and also expected outputs
+also in docstrings.
+The glue code for the "inner" features is separated below into "inner" and "outer" steps.
```
noodnik2, line 39: Huh??

internal/utils/utils.go
```diff
@@ -19,3 +20,100 @@ func S(n int) string {
 var TimeNowFunc = func() time.Time {
 	return time.Now()
 }
+
```
noodnik2, line 23: mega-`nit`:
- Consider creating a new file for your new formatter utilities - e.g., `formatters.go`
- Adding a new set of tools - especially if they focus on a single, new concern - into something as abstract a `utils.go` is a pet peeve of mine.

run.go
```diff
@@ -156,13 +157,16 @@ func (r *runner) concurrent(rate int) (failed bool) {
...
+	multiFmt, err := configureMultiFormatter(opt, output)
+	if err != nil {
+		return nil, err
 	}
 
-	multiFmt := ifmt.MultiFormatter{}
+	return multiFmt.FormatterFunc(suiteName, output), nil
+}
+
+func configureMultiFormatter(opt Options, output io.WriteCloser) (multiFmt ifmt.MultiFormatter, err error) {
```
noodnik2, line 169: `nit`: Consider dropping use of named return values (cost/benefit)?

noodnik2 (commented):
First, thanks for the effort to counter entropy - always a good & worthy battle if the heart is in the right place, and I "feel" it is here from what I've seen.  😊

I've sprinkled in a few "first impression" (mostly `nit` type) comments without having really grokked the changes, which were just too many to take in at once.  

Consider breaking this up into (maybe prioritized) sub-concerns?  I'd be more likely to be interested & able to give you better feedback & generate more excitement for pulling downstream. 🤓

internal/formatters/fmt_multi.go
```diff
@@ -34,6 +34,18 @@ func (r repeater) SetStorage(s *storage.Storage) {
...
+	var err error
+	for _, f := range r {
+		e := f.Close()
+		if e != nil {
+			err = e
+		}
+	}
+	return err
+}
+
```
nhv96, line 48: You forgot to update the comment. Also do we need to add test for this method?

run.go
Johnlon: agree

internal/formatters/fmt_multi.go
Johnlon: Will see what the existing test strat is for this and look to improve.

Johnlon:
> First, thanks for the effort to counter entropy - always a good & worthy battle if the heart is in the right place, and I "feel" it is here from what I've seen. 😊
> 
> I've sprinkled in a few "first impression" (mostly `nit` type) comments without having really grokked the changes, which were just too many to take in at once.
> 
> Consider breaking this up into (maybe prioritized) sub-concerns? I'd be more likely to be interested & able to give you better feedback & generate more excitement for pulling downstream. 🤓

I think this PR is already the atom of improvement.
I'd prefer to get it merged and then add the fixes for the bugs that motivated this

internal/utils/utils.go
Johnlon: will look

run.go
```diff
@@ -196,61 +196,140 @@ func runWithOptions(suiteName string, runner runner, opt Options) int {
...
+}
 
-	if opt.ShowStepDefinitions {
+// TestSuite allows for configuration
+// of the Test Suite Execution
+type TestSuite struct {
+	Name                 string
+	TestSuiteInitializer func(*TestSuiteContext)
+	ScenarioInitializer  func(*ScenarioContext)
+	Options              *Options // TODO mutable value - is this necessary?
```
nhv96, line 231: With your refactoring and move the function body to one upper level, I can only think the reason for us to keep using pointer here is to avoid incompatible with older version :thinking: 
Looks like they wanted to separate the concerns, test suite initialization vs runner logic to run that suite, so changed to the options configuration within the function body of runner should not reflected to the test suite.

Johnlon: yes I suspect you are right - how tolerant should we be of breaking changes.
I had undone some work to avoid breaks in this pr.

nhv96: I'd say we can tolerant to the point that user's expectations and use cases of `godog` library still remain the same. Except for the hook ordering bug we're going to fix, I really don't see we're going to "break" anything else.

internal/formatters/fmt_base.go
```diff
@@ -21,12 +21,12 @@ import (
...
+func BaseFormatterFunc(suite string, out io.WriteCloser) formatters.Formatter {
 	return NewBase(suite, out)
 }
 
 // NewBase creates a new base formatter.
-func NewBase(suite string, out io.Writer) *Base {
+func NewBase(suite string, out io.WriteCloser) *Base {
 	return &Base{
 		suiteName: suite,
 		indent:    2,
```
nhv96, line 32: Regarding this approach, do you think it will force the users (who use custom fmt) to working too much on their end to adapt to our changes? I think having a `Close()` method is a great enhancement, but I'm considering the case that someone is having problem with hook ordering bug, and expecting the new release will fix it, but then in order to use it they need to update their custom code first...

Should we use another way such as saying "we're going to fix this, and also we're migrating the Formatter API to a new version, you don't need to update anything yet and still have new features to use"? I have the following code that seems to work on my local testing
```suggestion
type baseAdapter struct {
	out io.Writer
}

func (f baseAdapter) Write(p []byte) (n int, err error) {
	return f.out.Write(p)
}

func (f baseAdapter) Close() error {
	return nil
}

// NewBase creates a new base formatter.
func NewBase(suite string, out io.Writer) *Base {
	var wc io.WriteCloser

	w, ok := out.(io.WriteCloser)
	if ok {
		wc = w
	} else {
		wc = baseAdapter{out: out}
	}

	return &Base{
		suiteName: suite,
		indent:    2,
		out:       wc,
		Lock:      new(sync.Mutex),
	}
}

// Base is a base formatter.
type Base struct {
	suiteName string
	out       io.WriteCloser
	indent    int

	Storage *storage.Storage
	Lock    *sync.Mutex
}
```

This way we can still migrate to use `io.WriteCloser` in our internal logic, but user don't need to know or don't need to update anything right away.
Ps: sorry my english is not that good so my reasoning is kinda long.

PR #668, opened 2024-11-27
Print step declaration line instead of handler declaration line

SpencerC:
### 🤔 What's changed?

Pretty printing results now prints the line where the step is declared instead of the line where the handler is declared.

Printing the file and line number of handlers often produces meaningless results. For example:
* Anonymous functions: `ctx.Step("...", func () {...})` always point to line 0: `some_file_test.go:0`
* Instance receiver methods: `ctx.Step("...", instance.Method)` always point to: `<autogenerated>:1`. 

Pointing to the step declaration does require an additional click to get to the handler code, so an alternative approach would be to improve the reflection process to do a better job locating the handler code.

This might be considered breaking if anyone is relying on the previous specific log format in CI tasks.

### 📋 Checklist:

----


vearutop:
Tested locally, looks good. 👍

PR #682, opened 2025-03-19
Enhancement: Support for custom types + `StepParam` type

tigh-latte:
### 🤔 What's changed?

Add a new supported type to step definitions, being:
- Any type that is structured ontop of another supported type:
```go
type MyCustomInt int
```
- any type that implements `StepParam`:
```go
type TestFile string

func (t *TestFile) LoadParam(ctx context.Context) (string, error) {}
```

We make fairly heavy use of files for test data, along with fairly heavy use of go's template engine for our inputs and file content. This, while working well, has resulted in basically every single step we define having to start with one or more of a few common function calls:

```go
func iExpectXToMatchY(ctx context.Context, x string, y string) error {
    rendered, err := loadAndRenderTestFile(y)
    if err != nil { /* you know how it goes */ }

    // assert that x == y
}
```

Which, while fine, is duplication that we've had to change a few times before. Combine this with not all templated input being files to be loaded, and you get a fair amount of duplicated boilerplate that eventually the `nodupl` linter complains about.

Ideally, this would be automatic while also being more declarative:

```go
func iExpectXToMatchY(ctx context.Context, x string, y TestTemplateFile) error {
    //assert that x == string(y)
}
```

Naming, new type placement, and general opinions on the feature would be good (such as, do we even want support for customer non-struct types?).

### 📋 Checklist:

----


vearutop (approved):
👍 

Looks good and useful to me!

internal/models/stepdef.go
```diff
@@ -230,6 +246,36 @@ func (sd *StepDefinition) Run(ctx context.Context) (context.Context, interface{}
...
+		return val, err
+	}
+
+	if param.Kind() == reflect.Ptr {
+		val = reflect.ValueOf(&s).Convert(param)
+	} else {
+		val = reflect.ValueOf(s).Convert(param)
+	}
+
+	tm := val.Interface().(StepParam)
```
vearutop, line 266: Is it possible for value to fail type assertion and panic here?

Issue #658, opened 2024-10-28
Retry failed tests automatically


harikamankala:
### 🤔 What's the problem you're trying to solve?

It is always frustrating when the test fails with no reason and just works fine with a retry. It is common to fail one or other test when there are thousands of tests running in parallel. There can be multiple reasons for failure like network slowness, overloaded server.

### ✨ What's your proposed solution?

It would be really helpful if we have a feature to rerun the failed tests.
Rerun option is also provided by other tools for example

ctest -> https://cmake.org/cmake/help/v3.17/manual/ctest.1.html ( --repeat until-pass:n )
maven-surefire-plugin -> https://maven.apache.org/surefire/maven-surefire-plugin/examples/rerun-failing-tests.html

I have also seen in here https://github.com/cucumber-rs/cucumber/issues/212

It would be really helpful, if you consider this. Thank you :)

nhv96:
I want to share my experience about test failing sometimes without reasons (aka flaky tests), and how I and my team solved it.
When using `godog` to apply BDD test for our product, which is built on top of microservices architecture, we have thousand of features and even though we tried so hard to keep each of them separated with each other in term of context, still there are many reason that can lead to the test fail, most common are:
- async read/write in DB layer: a step is supposed to insert data into a table, and another step is reading from it, but sometime the logic implementation of this feature is running concurrently, and the second step failed to read that data out, causing a fail.
- race condition in DB layer: as you can tell, maybe there are some other features happened to read/write into the same data in DB, causing unexpected behavior.
- async pubsub: a feature is supposed to test if service A can send a message to service B using Kafka, so first step is service A publish a message, second step is check for service B has received the message from Kafka. Sometime when the system running in high load, the message can arrive late in service B, causing the test failed.
And there're more...

But as users using the library to test, we did not expect anything more from `godog`, and the problem here is clearly that we didn't design our system good enough, we didn't thoroughly think of the real scenario when implement the logic,  our tests are not well prepared (should use hooks to setup DB, each feature or service should have it's own environment to run, etc...).
Thanks to flaky tests like this, we can know what problem is existing in our system, since it can occur in tests, it will definitely occur in production, and yup we did have some incidents and avoided some, thanks to `godog`.

What we did back then to avoid the flaky tests:
- Review again our system design of the feature (not the `.feature` files), try to over look or understand it fully, if there are many edge cases that can happen or risky, we refactor it.
[...]

PR #684, opened 2025-03-29
Implement retry flaky steps

nhv96:
### 🤔 What's changed?

- Added support for retrying flaky steps by returning `ErrRetry`
- Added CLI option for `MaxRetries` attempts

image

Discussed [here](https://github.com/cucumber/godog/issues/658).

### 📋 Checklist:

----

### DIFF
--- a/_examples/retry-flaky/features/retry.feature
+++ b/_examples/retry-flaky/features/retry.feature
@@ -1,0 +1,19 @@
+Feature: Godog should be able to retry flaky tests
+    In order to help Go developers deal with flaky tests
+    As a test suite
+    I need to be able to return godog.Err to mark which steps should be retry
+
+    Scenario: Test cases that pass aren't retried
+        Given a step that always passes
+
+    Scenario: Test cases that fail are retried if within the limit
+        Given a step that passes the second time
+
+    Scenario: Test cases that fail will continue to retry up to the limit
+        Given a step that passes the third time
+
+    Scenario: Test cases won't retry after failing more than the limit
+        Given a step that always fails
+
+    Scenario: Test cases won't retry when the status is UNDEFINED
+        Given a non-existent step
--- a/_examples/retry-flaky/godogs_test.go
+++ b/_examples/retry-flaky/godogs_test.go
@@ -1,0 +1,57 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"testing"
+
+	"github.com/cucumber/godog"
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_RetryFlaky(t *testing.T) {
+	suite := godog.TestSuite{
+		Name:                "retry flaky tests",
+		ScenarioInitializer: InitializeScenario,
+		Options: &godog.Options{
+			Format:     "pretty",
+			Paths:      []string{"features/retry.feature"},
+			MaxRetries: 2,
+		},
+	}
+
+	assert.Equal(t, 0, suite.Run())
+}
+
+func InitializeScenario(sc *godog.ScenarioContext) {
+	sc.Step(`^a step that always passes`, func(ctx context.Context) (context.Context, error) {
+		return ctx, nil
+	})
+
+	secondTimePass := 0
+	sc.Step(`^a step that passes the second time`, func(ctx context.Context) (context.Context, error) {
+		secondTimePass++
+		if secondTimePass < 2 {
+			return ctx, fmt.Errorf("unexpected network connection, %w", godog.ErrRetry)
+		}
+		return ctx, nil
+	})
+
+	thirdTimePass := 0
+	sc.Step(`^a step that passes the third time`, func(ctx context.Context) (context.Context, error) {
+		thirdTimePass++
+		if thirdTimePass < 3 {
+			return ctx, fmt.Errorf("unexpected network connection, %w", godog.ErrRetry)
+		}
+		return ctx, nil
+	})
+
+	fifthTimePass := 0
+	sc.Step(`^a step that always fails`, func(ctx context.Context) (context.Context, error) {
+		fifthTimePass++
+		if fifthTimePass < 5 {
+			return ctx, fmt.Errorf("must fail, %w", godog.ErrRetry)
+		}
+		return ctx, nil
+	})
+}
--- a/internal/flags/flags.go
+++ b/internal/flags/flags.go
@@ -46,4 +46,6 @@ built-in formatters are:
 specify SEED to reproduce the shuffling from a previous run
   --random=5738`)
 	flagSet.Lookup(prefix + "random").NoOptDefVal = "-1"
+
+	flagSet.IntVar(&opts.MaxRetries, prefix+"retry", opts.MaxRetries, "retry n times when a step encounter error")
 }
--- a/internal/flags/flags_test.go
+++ b/internal/flags/flags_test.go
@@ -25,6 +25,7 @@ func Test_BindFlagsShouldRespectFlagDefaults(t *testing.T) {
 	assert.False(t, opts.Strict)
 	assert.False(t, opts.NoColors)
 	assert.Equal(t, int64(0), opts.Randomize)
+	assert.Equal(t, 0, opts.MaxRetries)
 }
 
 func Test_BindFlagsShouldRespectFlagOverrides(t *testing.T) {
@@ -37,6 +38,7 @@ func Test_BindFlagsShouldRespectFlagOverrides(t *testing.T) {
 		Strict:              true,
 		NoColors:            true,
 		Randomize:           11,
+		MaxRetries:          1,
 	}
 	flagSet := pflag.FlagSet{}
 
@@ -51,6 +53,7 @@ func Test_BindFlagsShouldRespectFlagOverrides(t *testing.T) {
 		"--optOverrides.strict=false",
 		"--optOverrides.no-colors=false",
 		"--optOverrides.random=2",
+		"--optOverrides.retry=3",
 	})
 
 	assert.Equal(t, "junit", opts.Format)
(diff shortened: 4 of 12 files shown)


internal/flags/flags.go
```diff
@@ -46,4 +46,6 @@ built-in formatters are:
 specify SEED to reproduce the shuffling from a previous run
   --random=5738`)
 	flagSet.Lookup(prefix + "random").NoOptDefVal = "-1"
+
+	flagSet.IntVar(&opts.MaxRetries, prefix+"retry", opts.MaxRetries, "retry n times when a step encounter error")
```
luke-hill, line 50:
```suggestion
	flagSet.IntVar(&opts.MaxRetries, prefix+"retry", opts.MaxRetries, "Specify the number of times to retry failing tests (default: 0)")
```
Make sure this is defaulting to `0`

If you want to see where the original config parser is set with all of the messages check here: https://github.com/cucumber/cucumber-ruby/blob/main/lib/cucumber/cli/options.rb

internal/flags/flags_test.go
```diff
@@ -61,4 +64,5 @@ func Test_BindFlagsShouldRespectFlagOverrides(t *testing.T) {
 	assert.False(t, opts.Strict)
 	assert.False(t, opts.NoColors)
 	assert.Equal(t, int64(2), opts.Randomize)
+	assert.Equal(t, 3, opts.MaxRetries)
```
luke-hill, line 67: Check here also. This might need updating. Although given this is the same as the above maybe this is just asserting that?

luke-hill (changes requested):
Some minor fixes suggested. However a larger question is, are you seeking conformance with standard cucumber execution / cck conformance (I know you might not be there yet).

The reason why, is that a lot of your implementation details surrounding the specific status is unique and not what all standard implementors do

internal/flags/flags_test.go
```diff
@@ -51,6 +53,7 @@ func Test_BindFlagsShouldRespectFlagOverrides(t *testing.T) {
 		"--optOverrides.strict=false",
 		"--optOverrides.no-colors=false",
 		"--optOverrides.random=2",
+		"--optOverrides.retry=3",
```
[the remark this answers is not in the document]
nhv96: Ah this is not concurrency option. In this test MaxRetries is set to 1 when initialize the option object and then we pass in CLI flags to override the option value, the goal is to make sure that CLI options work as expected.

internal/flags/flags_test.go
nhv96: Same as above, this is to make sure the MaxRetries option can be assigned via CLI.

internal/flags/flags.go
nhv96: Yes it will be default 0. I'll update the message.

_examples/retry-flaky/godogs_test.go
```diff
@@ -0,0 +1,57 @@
...
+	thirdTimePass := 0
+	sc.Step(`^a step that passes the third time`, func(ctx context.Context) (context.Context, error) {
+		thirdTimePass++
+		if thirdTimePass < 3 {
+			return ctx, fmt.Errorf("unexpected network connection, %w", godog.ErrRetry)
+		}
+		return ctx, nil
+	})
+
+	fifthTimePass := 0
```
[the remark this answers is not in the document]
nhv96: At first it was written as you suggested, but then I thought we will need to cover the case that retry attempt runs out and eventually failed.
So can I add another step `a step that retry more than attemp limit`?

_examples/retry-flaky/godogs_test.go
```diff
@@ -0,0 +1,57 @@
...
+func InitializeScenario(sc *godog.ScenarioContext) {
+	sc.Step(`^a step that always passes`, func(ctx context.Context) (context.Context, error) {
+		return ctx, nil
+	})
+
+	secondTimePass := 0
+	sc.Step(`^a step that passes the second time`, func(ctx context.Context) (context.Context, error) {
+		secondTimePass++
+		if secondTimePass < 2 {
+			return ctx, fmt.Errorf("unexpected network connection, %w", godog.ErrRetry)
```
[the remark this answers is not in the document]
nhv96: Right, I think I should just return godog.ErrRetry instead.

nhv96:
> Some minor fixes suggested. However a larger question is, are you seeking conformance with standard cucumber execution / cck conformance (I know you might not be there yet).
> 
> The reason why, is that a lot of your implementation details surrounding the specific status is unique and not what all standard implementors do

@luke-hill I agree that this implementation works for the specific status `godog.ErrRetry`, in our discussion thread we also proposed other approaches as well such as using tags `@flaky` in the feature files.

Do you mean that other standard implementors will also implement such approaches as well? If it is the case, then this PR is to implement the approach using `RetryErr`, then the following PRs can be to add other approaches.

So I do want to conform the cucumber standard for implementations, but I don't want to make so many different changes in one PR :smile:

retry_flaky_test.go
```diff
@@ -0,0 +1,131 @@
...
+				thirdTimePass++
+				if thirdTimePass < 3 {
+					return ctx, fmt.Errorf("unexpected network connection, %w", ErrRetry)
+				}
+				return ctx, nil
+			})
+
+			fifthTimePass := 0
+			ctx.Step(`^a step that always fails`, func(ctx context.Context) (context.Context, error) {
+				fifthTimePass++
```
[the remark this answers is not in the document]
nhv96: Oh I forgot, this file is unit tests for the framework itself. So this case here is to cover the case step retried all attempts but still fail.

_examples/retry-flaky/godogs_test.go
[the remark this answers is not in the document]
nhv96: Yes I agree, this folder is for examples, I should not add this case here (this case is supposed to be in unit test over [here](https://github.com/cucumber/godog/pull/684#discussion_r2021180886) only).

nhv96:
hi @vearutop @luke-hill if you have time, please help me review this PR.

_examples/retry-flaky/features/retry.feature
```diff
@@ -0,0 +1,19 @@
+Feature: Godog should be able to retry flaky tests
+    In order to help Go developers deal with flaky tests
+    As a test suite
+    I need to be able to return godog.Err to mark which steps should be retry
```
luke-hill, line 4: more a question, but usually in feature files if it's a technical term you're referencing you'd use backticks or something else.

But maybe you don't do that here :man_shrugging:

_examples/retry-flaky/features/retry.feature
```diff
@@ -0,0 +1,19 @@
+Feature: Godog should be able to retry flaky tests
```
luke-hill, line 1: Whilst the principle here is good, I'd lean more on you implementing the CCK for things like this, so that you can use the conformance element there

internal/flags/options.go
```diff
@@ -80,6 +80,9 @@ type Options struct {
 
 	// ShowHelp enables suite to show CLI flags usage help and exit.
 	ShowHelp bool
+
+	// MaxRetries is the number of times you can retry failing tests. Default is 0 retry.
```
luke-hill, line 84:
```suggestion
	// MaxRetries is the number of times you can retry failing tests. Default is 0 retries.
```

internal/formatters/fmt.go
```diff
@@ -37,6 +37,7 @@ var (
 	undefined = models.Undefined
 	pending   = models.Pending
 	ambiguous = models.Ambiguous
+	retry     = models.Retry
```
luke-hill, line 40: Whilst this is godog, and you are allowed to do your own thing, if your ultimate goal is cucumber conformance, this would need to go.

We have a concept called "flaky" which is something that passed *after* retries, maybe this can be renamed or adapted I don't know

internal/models/results.go
```diff
@@ -74,6 +79,8 @@ const (
 	Pending
 	// Ambiguous ...
 	Ambiguous
+	// Retry ...
+	Retry
```
luke-hill, line 83: This might / should be altered too

internal/models/results.go
```diff
@@ -105,6 +112,8 @@ func (st StepResultStatus) String() string {
 		return "pending"
 	case Ambiguous:
 		return "ambiguous"
+	case Retry:
```
luke-hill, line 115: Here also

internal/storage/storage.go
```diff
@@ -336,3 +345,15 @@ func (s *Storage) mustGet(table, index string, args ...interface{}) memdb.Result
 
 	return it
 }
+
+func (s *Storage) first(table, index string, args ...interface{}) interface{} {
```
luke-hill, line 349: I am writing this comment to let you know I don't understand this bit. But I'm assuming it does something important :sweat_smile:

retry_flaky_test.go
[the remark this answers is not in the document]
luke-hill: So based on what I've seen earlier, this should ideally live in a separate folder called compatibility and you should look towards CCK conformance imo.

Maybe come and discuss this on discord. See what the desired avenue is in the go community

luke-hill (commented):
Done another pass at a review but at this point I think the "goal" needs to be understood better with regards to conformance / non-conformance

_examples/retry-flaky/features/retry.feature
nhv96: you're right, let me add it

nhv96:
hi @luke-hill sorry the my absent, I'm available to continue the work now :pray: i will ping you on discord

_examples/retry-flaky/features/retry.feature
luke-hill: So a potentially better way of getting all of this "conformance" is to use the CCK

https://github.com/cucumber/compatibility-kit

You can then aim for "partial" conformance, working your way to full conformance?

PR #686, opened 2025-06-28
chore: Bump go version to 1.22 and update dependencies - (#685)

### LINKED ISSUE #685: Update go.mod versions
## Context
The `go` version used in the project begins to be old *(using 1.16 while we are at the 1.24)* and dependencies haven't been updated for a while too.

I propose to update to 1.22, like this it still usable by many without requesting them to update there project.

## Objective
Update the `go` version to 1.22 and bump dependencies to the latest versions available for 1.22.

Adaendra:
### 🤔 What's changed?

Golang version has been updated to 1.22, breaking changes have been fixed, dependencies have been updated.

Fixes #685 

### 📋 Checklist:

### DIFF
--- a/internal/formatters/fmt_progress.go
+++ b/internal/formatters/fmt_progress.go
@@ -8,7 +8,7 @@ import (
 	"strings"
 
 	"github.com/cucumber/godog/formatters"
-	messages "github.com/cucumber/messages/go/v21"
+	messages "github.com/cucumber/messages/go/v24"
 )
 
 func init() {
--- a/formatters/fmt.go
+++ b/formatters/fmt.go
@@ -4,7 +4,7 @@ import (
 	"io"
 	"regexp"
 
-	messages "github.com/cucumber/messages/go/v21"
+	messages "github.com/cucumber/messages/go/v24"
 )
 
 type registeredFormatter struct {
--- a/go.mod
+++ b/go.mod
@@ -3,8 +3,8 @@ module github.com/cucumber/godog
 go 1.22
 
 require (
-	github.com/cucumber/gherkin/go/v26 v26.2.0
-	github.com/cucumber/messages/go/v21 v21.0.1
+	github.com/cucumber/gherkin/go/v32 v32.2.0
+	github.com/cucumber/messages/go/v24 v24.1.0
 	github.com/hashicorp/go-memdb v1.3.5
 	github.com/spf13/cobra v1.9.1
 	github.com/spf13/pflag v1.0.6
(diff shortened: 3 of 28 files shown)


.github/workflows/test.yml
```diff
@@ -8,7 +8,7 @@ jobs:
   test:
     strategy:
       matrix:
-        go-version: [ 1.16.x, 1.17.x, oldstable, stable ] # Lowest supported and current stable versions.
+        go-version: [ 1.22.x, 1.23.x, 1.24.x, oldstable, stable ] # Lowest supported and current stable versions.
```
vearutop, line 11: `1.23` and `1.24` are already here as `oldstable` and `stable`.

I think we can migrate to a list like `[<min_supported>, oldstable, stable]`, which would target the newest versions as main ones, but also will force backwards compatibility for users that lag behind.

I agree `go1.16` feels too old to support, although I would not be terribly surprised to still find enterprise `godog` users still stuck at 1.16.
 
Let's bump minimal supported version to 1.18 (3 years old now), wdyt?

internal/formatters/fmt_progress.go
```diff
@@ -42,7 +42,7 @@ func (f *Progress) Summary() {
 	left := math.Mod(float64(*f.Steps), float64(f.StepsPerRow))
 	if left != 0 {
 		if *f.Steps > f.StepsPerRow {
-			fmt.Fprintf(f.out, s(f.StepsPerRow-int(left))+fmt.Sprintf(" %d\n", *f.Steps))
+			fmt.Fprintf(f.out, "%s", s(f.StepsPerRow-int(left))+fmt.Sprintf(" %d\n", *f.Steps))
```
mpkorstanje, line 45: Would it make sense to replace the string concatenation with a format string like `%s %d\n` here?

mpkorstanje (commented):
Looks good to me. Small nitpick about format strings.

internal/builder/builder_vendored_test.go
Adaendra, line 1: This file has been added to test correctly the case where godog package has been vendored. As I need to stub some variables from the `builder` package, I needed to create a dedicated test file for it. Otherwise, I can't access these variables. *(The other test file being in the `builder_test` package.)*

.github/workflows/test.yml
vearutop: It is indeed a good practice to keep tech stack up to date, however it is not always feasible for large long-living projects. Forcing such users into an upgrade is a mild violation of backwards compatibility. It may still be an okay price to pay, given there are substantial advantages and clear justification.

For example, https://github.com/rs/zerolog/blob/master/go.mod still requires `go1.15`, even though this library receives regular updates. That's a friendly attitude towards users.

> Also I want to avoid the case where someone need to downgrade their version to be able to use a tool. (I had this case multiple times recently) So that's why it was for me a good compromise.

Could you share some context why the downgrade was needed?

mpkorstanje: Another aspect to consider is that Godog is an unfunded opensource project. The most important resource it has are its maintainers and contributors. Keeping the required go version somewhat recent lowers the barrier to entry for those - nobody likes working on really old stuff.

So I would suggest striking a balance. For example, the minimum required Go version could be the oldest support major Go release. That will impact users that don't keep their stack up to date. But unless they're paying for the privilege or are an active participants in the project, I don't see why that should be a problem for the participants in the Godog project.

> Forcing such users into an upgrade is a mild violation of backwards compatibility. It may still be an okay price to pay, given there are substantial advantages and clear justification.

Godog is currently on major version 0, so following Semver that would allow for any breaking changes at any moment. Users should expect breaking changes inside this major. If Godog were at v1, then changing the required go version would be a breaking change that should result in a new major release.

PR #692, opened 2025-07-15
fix: verify dogT exists in the context before using it

cakoolen:
getTestingT(ctx) can return a nil pointer (if there is no dogT in the context) in this case a segfault causes the whole test harness to fail. By verifying whether there is a dogT in the context, the test can continue.

### 🤔 What's changed?

Added a check to verify whether the dogT was available in the context before querying its state.

While using godog (After updating from v0.12.x to v0.15.0) our test harness ran into a segfault. The cause of the segfault was a nil pointer deference that is verified to be fixed with the proposed change.

### 📋 Checklist:

----

### DIFF
--- a/suite.go
+++ b/suite.go
@@ -159,7 +159,9 @@ func (s *suite) runStep(ctx context.Context, pickle *Scenario, step *Step, scena
 
 		// Check for any calls to Fail on dogT
 		if err == nil {
-			err = getTestingT(ctx).isFailed()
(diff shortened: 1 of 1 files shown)


vearutop (approved):

vearutop:
New version https://github.com/cucumber/godog/releases/tag/v0.15.1 released with this and a few other changes. Thank you for contribution!

cakoolen:
Thank you for merging it so quick and create the release!

PR #698, opened 2025-07-23
ci: use golangci lint

### LINKED ISSUE #696: 🔧 Integrate golangci-lint for Static Code Analysis
### 🤔 What's the problem you're trying to solve?

To improve code quality and maintain consistency across the project, I'd like to propose adding golangci-lint to our development workflow.

This tool helps catch common bugs, enforces style guidelines, and ensures adherence to Go best practices—all with minimal setup and configuration.

Why this matters:

- Identifies potential bugs early through static analysis
- Maintains consistent code style and formatting
- Reduces technical debt over time
- Speeds up code reviews with automated checks

Proposed steps:

- Add golangci-lint configuration file (.golangci.yml) to the root of the repository.
- Define rules and linters we want to enforce.
- Update documentation or contributing guide to include linting steps.
- Integrate it into the CI pipeline to auto-run checks on PRs.

Let me know your thoughts or if you'd like help setting up the initial config. 🚀

### ✨ What's your proposed solution?

Integrate the linter in the repository

### ⛏ Have you considered any alternatives or workarounds?

_No response_

### 📚 Any additional context?

_No response_

chengxilo:
### 🤔 What's changed?

1. replace the previous static check in workflow to golangci-lint
2. code was edited to meet the requirement of golangci-lint

Fix #696 
and my presonal preference.

### 📋 Checklist:

pull request.

----

### DIFF
--- a/cmd/godog/internal/cmd_root.go
+++ b/cmd/godog/internal/cmd_root.go
@@ -1,8 +1,10 @@
 package internal
 
 import (
+	"fmt"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
+	"log"
 
 	"github.com/cucumber/godog/internal/flags"
 )
@@ -23,7 +25,10 @@ and contain buildable go source.`,
 		RunE: runRootCmd,
 	}
 
-	bindRootCmdFlags(rootCmd.Flags())
+	err := bindRootCmdFlags(rootCmd.Flags())
+	if err != nil {
+		log.Println(err)
+	}
 
 	return rootCmd
 }
@@ -44,7 +49,7 @@ func runRootCmd(cmd *cobra.Command, args []string) error {
 	return runCmdRunFunc(cmd, args)
 }
 
-func bindRootCmdFlags(flagSet *pflag.FlagSet) {
+func bindRootCmdFlags(flagSet *pflag.FlagSet) error {
 	flagSet.StringVarP(&output, "output", "o", "", "compiles the test runner to the named file")
 	flagSet.BoolVar(&version, "version", false, "show current version")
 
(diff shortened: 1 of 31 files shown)


chengxilo:
I’ve noticed that the Codecov workflow fails on this pr while its changes don’t significantly affect code coverage. So I think for this pr maybe it is a good idea to igonore the Codecov report.

chengxilo:
Hi @Saurabh2402,

Thanks for your feedback!

Regarding the linters: here are the default ones used by golangci-lint – https://golangci-lint.run/usage/linters/#enabled-by-default. While I don't think this usually causes inconsistencies between local and CI environments, I agree that adding a configuration file would be a good idea if we want more control over which linters are enabled or disabled. Probably you could tell me which lint you want to enable, so I can add the configuration file with those lints enabled.

On the code quality concerns: I tend to use `_, _ = fmt.Fprintf()` when printing error messages (most of them are for error output), since there's often not much we can do if an error occurs at that point. That said, you're right—there are some error values I probably shouldn't ignore. I'll go through and fix those (for the non-error part).

Regarding handling `defer file.Close()`, you might be right—I’ll remove the redundant error checks if they’re not useful.

As for future PRs: if the code isn’t too terrible, maybe the best way would be to leave a review or you can feel free to create a new pr.

Thanks again!

cmd/godog/internal/cmd_root.go
```diff
@@ -23,7 +25,10 @@ and contain buildable go source.`,
 		RunE: runRootCmd,
 	}
 
-	bindRootCmdFlags(rootCmd.Flags())
+	err := bindRootCmdFlags(rootCmd.Flags())
+	if err != nil {
+		log.Println(err)
```
kcross-ctoken, line 30: proabaly should exit with error here

kcross-ctoken:
👍  .... I think we need golangci-lint

cmd/godog/internal/cmd_root.go
chengxilo: Yeah makes sense, when bindRootCmdFlags failed it should not be booted. I should use panic/Fatal instead ,will fix it

PR #701, opened 2025-07-30
fix: Validate the handler function param types when registering steps.

chengxilo:
### 🤔 What's changed?

It checks the step handler params while registering the steps.

There is a FIXME tag in test_context.go file and I think I can fix it.

I think we should have different error message when step registering/run because the `context.Context` param is in a wrong place. So probably it would be a good idea to fix it in another pr, wondering whether I should fix it in a new pr.

for example: 
when there is a handler like this:

```golang
func nokInvalidParamCtxNotFirst(_ int, _ context.Context){}
```

Maybe we can panic with

```golang
fmt.Sprintf("%s: the data type of parameter %d context.Context should be the first param", models.ErrUnsupportedParameterType, i)
```

### 📋 Checklist:

----

### DIFF
--- a/internal/models/stepdef.go
+++ b/internal/models/stepdef.go
@@ -12,7 +12,7 @@ import (
 	"github.com/cucumber/godog/formatters"
 )
 
-var typeOfBytes = reflect.TypeOf([]byte(nil))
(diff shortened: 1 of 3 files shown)


chengxilo:
ready to be merged.

PR #740, opened 2026-05-07
fix(tags): the `,` operator was not working as an `OR` operator in the tags given

kcross-ctoken:
### 🤔 What's changed?

Fix for the tagging `,` not working correctly as an `OR`/`||` operator as specified in https://docs.behat.org/en/v2.5/guides/6.cli.html#gherkin-filters that is mentioned in the code.

I would like to be able to have  `@environment&&@test, @environment&&@test2` as my tags work correctly
and select all tests that are tagged with `@environment&&@test || @environment&&@test2`

### 📋 Checklist:
- [ ] I agree to respect and uphold the [Cucumber Community Code of Conduct](https://github.com/cucumber/.github/tree/main?tab=coc-ov-file)

----

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ This document is formatted according to the principles of [Keep A CHANGELOG](htt
 - fix: change bang to being in README by @nahomEagleLion in https://github.com/cucumber/godog/pull/687
 - Mark junit test cases as skipped if no pickle step results available by @mrsheepuk in https://github.com/cucumber/godog/pull/597
 - Print step declaration line instead of handler declaration line by @SpencerC in https://github.com/cucumber/godog/pull/668
+- fix(tags): the , operator was not working as an OR operator in the tags by @kcross-ctoken in https://github.com/cucumber/godog/pull/740
 
 ## [v0.15.0]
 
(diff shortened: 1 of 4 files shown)


kcross-ctoken:
@mpkorstanje I would love to get this fix out soon as its impacting my use of godog.... Otherwise Im going to have to run off a fork.

kcross-ctoken:
Ive fixed the mod in _examples....  Please re-approve for a ci run @mpkorstanje 😄 
Its quite weird _examples uses 1.23 and the main mod is 1.16 🤷 
Also I would recommend using golangci-lint instead of the `go run honnef.co/go/tools/cmd/staticcheck` in the make file ... I had to comment that out to run the tests due to to mod version issues 🤷 
Something like `asdf` or `mise` would help with standardising the tools to build and test.

_examples/go.mod
```diff
@@ -19,7 +19,7 @@ require (
 	github.com/davecgh/go-spew v1.1.1 // indirect
 	github.com/gofrs/uuid v4.3.1+incompatible // indirect
 	github.com/hashicorp/go-immutable-radix v1.3.1 // indirect
-	github.com/hashicorp/go-memdb v1.3.4 // indirect
+	github.com/hashicorp/go-memdb v1.3.5 // indirect
```
kcross-ctoken, line 22: not entirely sure where this came from ... I suspect the last commit to main did not do a `go mod tidy`

vearutop: Yeah, this is unrelated to your PR, there were some module issues in `main`, please rebase or merge main.

vearutop: `_examples` is sitting on the newer Go version because of its dependencies, main module tries to keep version requirement lower to not enforce upgrades on legacy users.

kcross-ctoken:
@vearutop just rebased ... ready to run though ci again.

vearutop (approved):
Looks good, this may come as unexpected behavior change if someone was using AND(OR...) precedence, and now it becomes OR(AND...), but since `godog` puts `behat` as inspiration, tags alignment can indeed be considered a fix of incorrect behavior.

PR #760, opened 2026-07-10
fix(hooks): wrap errors to preserve sentinel values

victoriacheng15:
### 🤔 What's changed?

This PR fixes hook error wrapping to preserve the types of all merged errors. 

Specifically:
*   Introduces a custom `joinedError` type and `mergeErrors` helper at the end of [suite.go](./suite.go).
*   Updates `runBeforeStepHooks`, `runAfterStepHooks`, `runBeforeScenarioHooks`, and `runAfterScenarioHooks` in [suite.go](./suite.go) to use `mergeErrors` instead of `fmt.Errorf("%v, %w")`.
*   This ensures that all errors returned from hook handlers are correctly wrapped, making them fully detectable by `errors.Is` and `errors.As`.

**Backward Compatibility**:
*   **Go 1.18 - 1.19**: Supported via the custom `Is` and `As` methods on `joinedError`, which have been natively supported by the standard `errors` package since Go 1.13.
*   **Go 1.20+**: Supported via the `Unwrap() []error` method, allowing modern Go runtimes and static analysis tools to natively unwrap the error slice.
*   **Error Messages**: Preserves the original `%v, %v` string formatting, ensuring no breaking changes to logs or output formatting.

Fixes #732

Currently, when a step fails and a hook also returns an error (such as a skip), they are merged using `fmt.Errorf("%v, %w")`. This prevents `errors.Is` and `errors.As` from detecting the hook error type because it is formatted with `%v` instead of `%w`. As a result, the runner fails to recognize sentinel errors like `ErrSkip` returned from hooks when a step fails.

Feedback on the `joinedError` implementation for Go versions < 1.20 (which do not support multiple `%w` verbs).

### 📋 Checklist:

----

### DIFF
--- a/suite.go
+++ b/suite.go
@@ -304,7 +304,7 @@ func (s *suite) runBeforeStepHooks(ctx context.Context, step *Step, err error) (
 			if err == nil {
 				err = herr
 			} else {
-				err = fmt.Errorf("%v, %w", herr, err)
+				err = mergeErrors(herr, err)
 			}
 		}
 
@@ -329,7 +329,7 @@ func (s *suite) runAfterStepHooks(ctx context.Context, step *Step, status StepRe
 			if err == nil {
 				err = herr
 			} else {
-				err = fmt.Errorf("%v, %w", herr, err)
+				err = mergeErrors(herr, err)
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


vearutop:
Looks good, thank you! Could you also add a test to avoid regressions in the future?

victoriacheng15:
@vearutop Done! Added the regression test `TestSuite_HookErrorWrapping` to suite_context_test.go and verified it passs locally under go 1.18 and 1.22.

vearutop (approved):

